PR #73, opened 2019-10-14
add Windows scripts

michelou:
Add preliminary support for running GraalSqueak on Windows:
- Windows command script `scripts/template.graalsqueak.cmd` is adapted from [`scripts/template.graalsqueak.sh`](https://github.com/hpi-swa/graalsqueak/blob/master/scripts/template.graalsqueak.sh).
- Windows batch script `scripts/make_component.bat`is adapted from [`scripts\make_component.sh`](https://github.com/hpi-swa/graalsqueak/blob/master/scripts/make_component.sh).

### DIFF
--- a/scripts/make_component.bat
+++ b/scripts/make_component.bat
@@ -1,0 +1,95 @@
+:: #
+:: # Copyright (c) 2019-2019 Software Architecture Group, Hasso Plattner Institute
+:: #
+:: # Licensed under the MIT License.
+:: #
+@echo off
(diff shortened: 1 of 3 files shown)


fniephaus:
Thanks a lot for the contribution! I would prefer to delay the integration of `make_component.bat` as it's currently unclear how we have to build GraalVM components for Windows. But more than happy to merge the rest once I verified everything works as intended.

PR #75, opened 2019-10-20
add Windows scripts (cont'd)

michelou:
Script file `make_component.sh` is missing one operation: add symlink entry for command file `graalsqueak.cmd`.

**CAUTION:**  Both script files `make_component.sh` and `make_component.bat` **MUST** be kept in sync. One trivial example is line 11: `readonly GRAALVM_VERSION="19.2.1"`.


michelou:
@fniephaus Please wait with merging this PR (sorry).
After reviewing my script file `build install` I think the component installation can be done without this change on Windows (I need to look again at the other third-party components).

michelou:
Fixed both files `scripts/make_component.bat` and `scripts/make_component.sh`.

PR #79, opened 2019-10-24
Sista and primitive calls

fmateoc:
Sorry about the pull request contents, I only meant to include the last commit, but I don't know how to achieve that on an ongoing basis. If you have a recommended workflow, please share. Also, is there a mailing list or forum for this project?

The commit is fixing some incorrect assumptions about Sista inlined primitive calls - they do not hard return - so they need a different behavior. As such, the fact that the primitive calls were left inside the bytecode loop did not address the issue, anyway the branch dealing with primitive calls within the loop only worked for normal primitives (it fetched from hardcoded index 0, it asserted successor index as 3, and it hard returned). Also, at primitive call node creation time, the primitive index was incorrect for Sista primitives. I separated the two kinds of primitive call nodes and moved the handling of the "normal" one outside the loop.
Also re-arranged the bytecode loop to do the tests in an order closer to descending probabilities


src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -210,6 +265,14 @@ private Object startBytecode(final VirtualFrame frame) {
                 pc = node.getSuccessorIndex();
                 if (node instanceof AbstractSendNode) {
                     FrameAccess.setInstructionPointer(frame, code, pc);
+                } else if (node instanceof SistaCallInlinePrimitiveNode) {
+                    scipProfile.enter();
```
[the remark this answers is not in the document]
fmateoc: I am new to truffle, so this may not do what I think it does
I intended to signal the compiler that this is an unlikely branch (as of now completely unused, as you pointed out)

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -210,6 +265,14 @@ private Object startBytecode(final VirtualFrame frame) {
...
+                    final SistaCallInlinePrimitiveNode callPrimitiveNode = (SistaCallInlinePrimitiveNode) node;
+                    if (callPrimitiveNode.primitiveNode != null) {
+                        // cannot fail
+                        callPrimitiveNode.primitiveNode.executePrimitive(frame);
```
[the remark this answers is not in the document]
fmateoc: oops, you are right, it should be pushed onto the stack

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -241,32 +296,32 @@ private Object resumeBytecode(final VirtualFrame frame, final long initialPC) {
...
                 pushClosureNode.executePush(frame);
                 pc = pushClosureNode.getClosureSuccessorIndex();
                 continue bytecode_loop_slow;
             } else {
-                final int successor = node.getSuccessorIndex();
+                pc = node.getSuccessorIndex();
                 if (node instanceof AbstractSendNode) {
-                    FrameAccess.setInstructionPointer(frame, code, successor);
+                    FrameAccess.setInstructionPointer(frame, code, pc);
+                } else if (node instanceof SistaCallInlinePrimitiveNode) {
```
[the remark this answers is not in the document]
fmateoc: I think this is an oops again, sorry

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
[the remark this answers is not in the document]
fniephaus: No problem at all, happy to help :)

fmateoc:
I should also add that, since performance is a very important goal here, I try to be very sensitive about it and proceed incrementally, checking performance at each step, that's why I went ahead with a pull request even though this is just a preliminary step

fmateoc:
And one more thing :)
I think some of the CI tests may be incorrect - I hit one of them the first time I pushed the commit, it resumed a primitive method from 0. Since there is no preemption point at 0 for a primitive method, its context cannot be suspended at 0, so I had assumed that this was impossible. I changed the code to make it work, but perhaps the test should have been fixed instead.
Also, unfortunately I could not make sense of the current failures

fniephaus:
> As a general comment: I was looking at the code trying to learn and because I want to try to implement preemption, which will need to touch this code - by the way, I think you mentioned having experimented with it - is there a branch that has the latest attempt?

I've tried a few things and I think some of it is on the [`pc-modifications` branch](https://github.com/hpi-swa/graalsqueak/tree/wip/pc-modifications). Modifying the PC, however, is performance critical. I think the best way to do this is do work with a Truffle assumption that assumes the context pc has not been modified. As long as the assumption holds, the current version of the interpreter loop should be used. Otherwise, we need to fall back to another (slower) version which always reads the pc from the context. This is easier said than done, which is why it's not yet implemented. It might not be a good way to start with GraalSqueak to be honest. There are still a lot of plugins/primitives that may be easier to start with.


> I wanted to reorganize the code a little in preparation for that (in particular getting the primitive handling code out of the loop) and I stumbled upon your comment that you left the primitive handling code inside the loop because Sista primitives may appear in the middle of a method. This made me curious and I looked at the current Squeak image, and read the EncoderForSistaV1 class comments, as well as at the InstructionStream>>interpretNext3ByteSistaV1Instruction:for:extA:extB:startPC: and the Context>>callPrimitive: methods.
So, even though the Sista bytecode set is currently not supported, I wanted to address your comment while also fixing some obvious inconsistencies - but I have to say that I do not have practical experience with Sista

I know a few things about Sista, and I tried to keep them in mind, so it will be easier to support it in the future. However, I would prefer to wait with integration something like a Sista-specific bytecode until we actually work towards supporting it. Implementing something that cannot be tested at the moment does not make much sense in the context of GraalSqueak. I hope you agree!

> I should also add that, since performance is a very important goal here, I try to be very sensitive about it and proceed incrementally, checking performance at each step, that's why I went ahead with a pull request even though this is just a preliminary step

That's good. We actually have a benchmark server set up which runs the Are We Fast Yet benchmarks on `dev` commits.  Also, it's worth looking at compiler graphs and Truffle logs every now and then.

> And one more thing :)
I think some of the CI tests may be incorrect - I hit one of them the first time I pushed the commit, it resumed a primitive method from 0. Since there is no preemption point at 0 for a primitive method, its context cannot be suspended at 0, so I had assumed that this was impossible. I changed the code to make it work, but perhaps the test should have been fixed instead.
Also, unfortunately I could not make sense of the current failures

Which of the CI tests do you think are incorrect and why? You are right that a context cannot be preempted at the first bytecode in case of a primitive method. However, it is possible that "fresh" contexts reach the resume context node at the moment. Making the scheduler work correctly was quite hard and the code is still a bit messy and not well documented at this point. Again, this might not be a good starting point to be honest.

fmateoc:
Thank you. Is there a way to run the CI tests locally?

fmateoc:
I mean within Eclipse

fmateoc:
Anyway, the failing test, which I think was incorrect, was SqueakBytecodeTest.testCallPrimitive(), but I assume that was just the first primitive test encountered, it would have likely failed in the same way for all senders of AbstractSqueakTestCase.runPrim(...)

fniephaus:
> Thank you. Is there a way to run the CI tests locally?

Yes, you can run all tests in `de.hpi.swa.graal.squeak.test` with Eclipse's JUnit integration:

image


> Anyway, the failing test, which I think was incorrect, was SqueakBytecodeTest.testCallPrimitive(), but I assume that was just the first primitive test encountered, it would have likely failed in the same way for all senders of AbstractSqueakTestCase.runPrim(...)

The bytecode loop is crafted very carefully and it's easy to break all kinds of things (including performance) when modifying it. So I'm not surprised you see a weird combination of failures. I might have time to look into this myself next week.

fmateoc:
Thanks, I figured out why my tests were not running - I did not add -XX:-UseJVMCIClassLoader to the testsuite configuration, I just tried to run them directly.
Now that I did, I can see that indeed, the test is creating a top level context with the primitive method, and then it tries to resume it from the start, which should not be possible

fmateoc:
And the new code caught one more invalid test: SqueakMiscellaneousTest.testSourceAllBytecodes(), which creates a method with a standard primitive call bytecode in the middle of the method.
I would argue that even if only for catching these invalid tests and the separation of the two kinds of primitive calls, and the making explicit that normal primitive calls can only be at index 0 was worth it

fniephaus:
I agree. If you could fix the test, that'd be great. Otherwise, I'll have a look soon.

fmateoc:
For now I updated the commit with the latest cleanups, and removing the workaround for resuming a primitive context at 0, replacing it with an assertion.
I will try to also fix the failing tests, at least the ones that I understand :)
If you prefer, I can also push a version of the code without the new inlined primitive call node, and removing the branch checking for it in the bytecode loop. I agree that it is forward looking and currently untestable

fniephaus:
Thanks for the update. I checked and the branch profile is not needed due to the way merge explode works (see the explode loop annotation). And let's turn the inline primitive into a (short) comment somewhere in the loop, so we know what to do when we work on support for Sista.

fmateoc:
I fixed all the failures that I encountered locally, but I don't have your test images, so those image-based tests did not run for me. I see that some are failing on the server, can you please take a look?
One of the issues I encountered was quite "interesting", trying to print a Squeak stack trace without an image, so no class objects were initialized - I wonder how that could have ever worked

fniephaus:
To download the test images, you need to run this:  https://github.com/hpi-swa/graalsqueak/blob/master/scripts/download-test-image.sh

Then, you should be able to run the SUnit tests.

Stack printing does not work for tests that run without an image, so it's expected that this breaks. We mainly test against all SUnit tests where printing the stack can help to understand an error.

fmateoc:
There is something dubious going on. I did not get that CI test error in my local runs, but: 
first of all, consecutive runs with no code changes have different results (failures), 
second, I did see among the failures some similarly dubious DNUs (which do not occur when run in Smalltalk), but they vanish when I run the same test individually. This leads me to believe they are the result of some optimizations gone astray (or a lack of some necessary invalidations), but I don't know how to go about debugging that.

fmateoc:
Things got a little better with the latest fix, but travis still gives an error that I cannot reproduce locally, whether run as a whole testuite or individually

fmateoc:
But I have to wonder what is travis really testing: how come it did not catch all those tests that were failing because of an attempt to start a process at priority 9? It also did not catch many others that were failing because of the bug with the utc tine vs local (I had also found it and fixed it while debugging the tests)

fniephaus:
> There is something dubious going on. I did not get that CI test error in my local runs, but:
first of all, consecutive runs with no code changes have different results (failures),
second, I did see among the failures some similarly dubious DNUs (which do not occur when run in Smalltalk), but they vanish when I run the same test individually. This leads me to believe they are the result of some optimizations gone astray (or a lack of some necessary invalidations), but I don't know how to go about debugging that.

Sounds like you're experiencing one of the many [pitfalls](https://stefan-marr.de/2018/04/a-truffle-specialization-pitfall/) that come with Truffle.

The [`.travis.yml`](https://github.com/hpi-swa/graalsqueak/blob/master/.travis.yml) contains step by step instructions for how CI runs the tests.  And indeed, it's quite hard to debug scheduling problems and misbehaving nodes...I might have some time to look into your issues later today. I'm sure it's easy to explain, but it's probably hard to track down the root cause.

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
```diff
@@ -77,6 +78,22 @@ private static void patchImageForTesting() {
...
+        final PointersObject priority40List = (PointersObject) lists.getObjectStorage()[USER_PRIORITY_LIST_INDEX];
+        final Object firstLink = priority40List.instVarAt0Slow(LINKED_LIST.FIRST_LINK);
+        priority40List.instVarAtPut0Slow(LINKED_LIST.FIRST_LINK, newProcess);
+        if (firstLink == NilObject.SINGLETON) {
+            priority40List.instVarAtPut0Slow(LINKED_LIST.LAST_LINK, newProcess);
+        } else {
+            newProcess.instVarAtPut0Slow(PROCESS.NEXT_LINK, firstLink);
+        }
+        newProcess.instVarAtPut0Slow(PROCESS.LIST, priority40List);
+    }
```
fniephaus: Why do you create a new process here for testing? More importantly, the process created here is incomplete (it doesn't have a `suspendedContext` for example), so you'll get an error in case this process is ever selected by the scheduler.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -58,11 +60,56 @@
...
+                    return callPrimitiveNode.primitiveNode.executePrimitive(frame);
+                } catch (final PrimitiveFailed e) {
+                    getHandlePrimitiveFailedNode().executeHandle(frame, e.getReasonCode());
+                    LOG.log(Level.FINE, () -> callPrimitiveNode.primitiveNode +
+                                    " (" + ArrayUtils.toJoinedString(", ", FrameAccess.getReceiverAndArguments(frame)) + ")");
+                    /* continue with fallback code. */
+                }
+            }
+            assert callPrimitiveNode.getSuccessorIndex() == CallPrimitiveNode.NUM_BYTECODES;
+            return super.executeFresh(frame);
```
fniephaus: Subclassing is potentially dangerous in Truffle. `ExecuteContextNode#executeFresh` is no longer final, so the compiler might use an invoke here if you are unlucky.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/WakeHighestPriorityNode.java
```diff
@@ -41,16 +43,18 @@ public void executeWake(final VirtualFrame frame) {
...
+                    throw SqueakException.create("Should not be reached");
+                }
             }
-            processList = (PointersObject) arrayReadNode.execute(schedLists, p--);
-        } while (processList.isEmptyList(pointersReadNode));
-        final PointersObject newProcess = processList.removeFirstLinkOfList(pointersReadNode, pointersWriteNode);
-        contextNode.executeGet(frame).transferTo(pointersReadNode, pointersWriteNode, newProcess);
+        }
+        errorProfile.enter();
+        throw SqueakException.create("scheduler could not find a runnable process");
```
fniephaus: I cannot tell whether the logic has changed here at all, or if you only replaced the `do{} while()` here with a `for` loop. In case you're wondering, we used `do` to be closer to the implementation in Slang and [SqueakJS](https://github.com/bertfreudenberg/SqueakJS/blob/96edc18e3fcaf4e637f2978b8f0b6c3c608d961f/vm.js#L5796-L5807).

fmateoc: You are right, I was mislead by some code in the image, ProcessorScheduler>>nextReadyProcess (used by the profiler), where if the first element happens to not have a suspended context, it is skipped and the iteration continues.
I should have checked Slang instead

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
fmateoc: I saw some errors in the tests (I think it must have been DebuggerTests>>test04.. and test05... where new processes are created and set to priority Processor activeProcess priority - 1. They were failing because the priority was 9, below the lowestPriority. The only process at 10 is the idleLoop, so I assumed that the active process for the doItContext was the idleLoop. Which could happen if the image was not reloaded. But I did miss setting my new process to be the active process. Anyway, this should be done only if the active process at the beginning of the evaluation is the idleLoop

fmateoc: Regarding your general comment, I agree, I got carried away - initially I just wanted to get the initial PR to pass the tests, but as I started to look at all those tests and discovering new things, I kept adding to it.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/WakeHighestPriorityNode.java
fmateoc: Actually, now that I did read what Slang does (in StackInterpreter>>wakeHighestPriority), the change was actually correct:
There is an explicit comment in the loop
			"Only answer processes with a runnable suspendedContext.
			  Discard those that aren't; the VM would crash otherwise."
...and then there is a check for a live context
			 (self isLiveContext: ctxt) ifTrue:
which was missing

fniephaus: Ok, great! Thanks for double checking.

fmateoc:
This latest bunch is not an official submission, I just had several more bugfixes and I was curious to see if CI would still complain or not, but no such luck, now there are some minor conflicts so CI does not even try

fmateoc:
One more try after merging with the latest from upstream and fixing a couple more issues. Still hoping for a successful CI test

fmateoc:
Sure, you may cherry-pick as you wish. I can also try to organize things a bit and create some clean pull requests, although I am not very good at git, I am not a command line person. I think my next project will be a Git UI client :)

fmateoc:
I think something is wrong with the CI tests. E.g. windows should also be included as a test platform, since it gives wildly different results and it has allowed me to find a fix a lot of different clear bugs that were not discovered with the current setup.
After trying for a long time to clear CI just by looking at the build results, I think I am ready to give up.
I just did a small experiment to confirm I was not going crazy and the bugs I fixed were real: I switched to the upstream dev branch and ran all the tests locally on my Windows machine with graalvm-ee-19.2.1. The tests completely broke down after around 3200 tests, the last 1830 ones all errorred out. As such, it finished early, in 2034s (essentially the last 1830 tests were skipped), anyway the results were 5236/5236 (354 skipped) Errors: 1836 Failures: 62. I have attached the results
[upstream.txt](https://github.com/hpi-swa/graalsqueak/files/3833202/upstream.txt)

fniephaus:
Hi @fmateoc,
Thanks for the update and sorry for the poor experience on Windows. It's the least supported platform at the moment, even GraalVM does not fully support it at this point. However, we appreciate any Windows-related bug reports and contributions. In #86, I suggested that we need a Windows CI build for building the pre-compiled component. I agree it'd also make sense to run the tests against Windows. Maybe you have some cycles to add a Windows build job to Travis CI?

RE testing pipeline breaking down: I've observed this behavior quite a lot. Sometime one test fails and leaves the image in an inconsistent state and then, all subsequent tests may fail as well. Your error log, for example, contains a lot of "Multi threaded access requested by thread" errors. Those are definitely triggered by the testing infrastructure.

fmateoc:
yes, and that's one of the things that I fixed, i.e. both cleaning up the image so that subsequent tests can run and making sure the worker thread of a timed out tests is terminated before continuing - bad things happen when two worker threads are running on the same image :)

fniephaus:
I've started to go over your individual commits and will try to cherry pick as many as I can. Thanks!

fmateoc:
I just went and looked at https://docs.travis-ci.com/user/reference/windows/ to see what needs to be done to add a Windows build job to Travis CI, and they don't list java as a supported language :(

fmateoc:
...but there might be some workarounds - I'll try

fniephaus:
I am not surprised considering their Windows integration is still in beta. Nonetheless, I don't think this is a big deal: GraalSqueak needs a special JDK/GraalVM anyway, so a pre-installed JDK wouldn't help. Might be harder to get Python installed on the machine to run the mx tool.

fmateoc:
I realized that there are also problems with the test image on Windows: essentially all the tests loaded with OSProcess are Unix-only, and they fail on Windows,. While some of them are ignored, the UnixProcessAccessor... ones are all failing or raising errors on Windows, but most are marked as passing in the test - well, this is a problem with the tests in GraalSqueak, but I suppose they could also be avoided by unloading them from the image used on Windows.
I also noticed that the test image did not initialize StackInterpreter's PrimitiveTable, which leads to many Decompiler failures, even though most of them succeed in Squeak.. While in Squeak the image also starts off with the PrimitiveTable uninitialized, it gets initialized by simulation tests that happen to run before the decompiler ones. Since we reload the image multiple times while running the testsuite, we lose that initialization, so we should have in the the image

fniephaus:
> I realized that there are also problems with the test image on Windows: essentially all the tests loaded with OSProcess are Unix-only, and they fail on Windows,. While some of them are ignored, the UnixProcessAccessor... ones are all failing or raising errors on Windows, but most are marked as passing in the test - well, this is a problem with the tests in GraalSqueak, but I suppose they could also be avoided by unloading them from the image used on Windows.

We might want to add and use some platform-specific flags, such as `IGNORED_WINDOWS`, in [`SqueakTests.java`](https://github.com/hpi-swa/graalsqueak/blob/ff919dc054c4bc371c18cb767e38f21b787ce37e/src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/SqueakTests.java#L35) to be able to exclude certain tests.

> I also noticed that the test image did not initialize StackInterpreter's PrimitiveTable, which leads to many Decompiler failures, even though most of them succeed in Squeak.. While in Squeak the image also starts off with the PrimitiveTable uninitialized, it gets initialized by simulation tests that happen to run before the decompiler ones. Since we reload the image multiple times while running the testsuite, we lose that initialization, so we should have it in the initial image

Interesting finding. [This is more or less](https://github.com/hpi-swa/graalsqueak/blob/ff919dc054c4bc371c18cb767e38f21b787ce37e/docs/TestImage.md) how the test image is created. I'd like the number of GraalSqueak-specific changes to be as small as possible, maybe the table can be initialized in a cleanup routine as part of [`SqueakSUnitTest.java`](https://github.com/hpi-swa/graalsqueak/blob/ff919dc054c4bc371c18cb767e38f21b787ce37e/src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/SqueakSUnitTest.java). I will have a look unless you beat me to it ;)

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/CompiledCodeObject.java
```diff
@@ -107,7 +108,15 @@ private void setLiteralsAndBytes(final Object[] literals, final byte[] bytes) {
 
     public final Source getSource() {
         if (source == null) {
-            source = Source.newBuilder(SqueakLanguageConfig.ID, CompiledCodeObjectPrinter.getString(this), toString()).build();
+            String contents, toString;
+            try {
+                contents = CompiledCodeObjectPrinter.getString(this);
+                toString = toString();
+            } catch (final RuntimeException e) {
```
fniephaus: In what situation can a `RuntimeException` be thrown here?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/RemoveProcessFromListNode.java
```diff
@@ -32,10 +32,12 @@ public final void executeRemove(final PointersObject process, final PointersObje
...
     @Specialization(guards = "process == first")
     protected final void doRemoveEqual(final PointersObject process, final PointersObject list, @SuppressWarnings("unused") final PointersObject first, final AbstractSqueakObject last) {
-        final Object next = readNode.execute(process, PROCESS.NEXT_LINK);
-        writeNode.execute(list, LINKED_LIST.FIRST_LINK, next);
         if (process == last) {
             writeNode.executeNil(list, LINKED_LIST.LAST_LINK);
+            writeNode.executeNil(list, LINKED_LIST.FIRST_LINK);
+        } else {
+            final Object next = readNode.execute(process, PROCESS.NEXT_LINK);
+            writeNode.execute(list, LINKED_LIST.FIRST_LINK, next);
```
fniephaus: When is this an issue? And why isn't it a problem in the reference VM? This is the corresponding Slang code:
```smalltalk
[nextLink := objectMemory fetchPointer: NextLinkIndex ofObject: aProcess.
 self deny: (objectMemory isForwarded: nextLink).
 objectMemory storePointer: FirstLinkIndex ofObject: aList withValue: nextLink.
 aProcess  = lastLink ifTrue:
	[objectMemory storePointerUnchecked: LastLinkIndex ofObject: aList withValue: objectMemory nilObject]]
```

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ResumeContextNode.java
```diff
@@ -41,6 +46,20 @@ protected final Object doResumeInMiddle(final ContextObject context) {
...
+    }
+
+    private SendSelectorNode getCannotReturnNode() {
+        if (cannotReturnNode == null) {
+            CompilerDirectives.transferToInterpreterAndInvalidate();
+            cannotReturnNode = insert(SendSelectorNode.create(executeContextNode.code, executeContextNode.code.image.cannotReturn));
+        }
+        return cannotReturnNode;
+    }
+
```
fniephaus: Can you provide some code that tests or makes use of this?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ResumeContextNode.java
```diff
@@ -41,6 +46,20 @@ protected final Object doResumeInMiddle(final ContextObject context) {
         return executeContextNode.executeResumeInMiddle(context.getTruffleFrame(), initialPC);
     }
 
+    @Fallback
```
fniephaus: I'm going to replace `@Fallback` with:
`@Specialization(guards = "context.getInstructionPointerForBytecodeLoop() < 0")`.

The generated fallback guard would currently check  the PC value twice (`== 0` and then `> 0`).

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/ContextObject.java
```diff
@@ -583,7 +601,9 @@ public FrameMarker getFrameMarker() {
 
     // The context represents primitive call which needs to be skipped when unwinding call stack.
     public boolean isPrimitiveContext() {
-        return !hasClosure() && getMethod().hasPrimitive() && getInstructionPointerForBytecodeLoop() <= CallPrimitiveNode.NUM_BYTECODES;
+        return !hasClosure() && getMethod().hasPrimitive() &&
+                        !getMethod().isUnwindMarked() && !getMethod().isExceptionHandlerMarked() &&
```
fniephaus: Are these two additional checks really necessary?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/HandleNonLocalReturnNode.java
```diff
@@ -33,7 +34,13 @@ public static HandleNonLocalReturnNode create(final CompiledCodeObject code) {
     protected final Object handleModifiedSender(final VirtualFrame frame, final NonLocalReturn nlr) {
         aboutToReturnNode.executeAboutToReturn(frame, nlr); // handle ensure: or ifCurtailed:
         final ContextObject newSender = FrameAccess.getSenderContext(frame); // sender has changed
-        final ContextObject target = (ContextObject) nlr.getTargetContextOrMarker();
```
fniephaus: Have you even seen this type cast fail?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/ResumeProcessNode.java
```diff
@@ -32,20 +34,29 @@ public static ResumeProcessNode create(final CompiledCodeObject code) {
...
 
-    @Specialization(guards = "!hasHigherPriority(newProcess)")
+    @Specialization(guards = {"!hasHigherPriority(newProcess)", "hasSuspendedContext(newProcess)"})
     protected final void doSleep(final PointersObject newProcess) {
         putToSleepNode.executePutToSleep(newProcess);
     }
 
+    @Specialization(guards = "!hasSuspendedContext(newProcess)")
+    protected final void doFail(@SuppressWarnings("unused") final PointersObject newProcess) {
+        throw PrimitiveFailed.GENERIC_ERROR;
```
fniephaus: Raising a `PrimitiveFailed` in a non-primitive node is potentially dangerous, but I think this node is always part of an `AbstractPrimitiveNode`. Nonetheless, could you provide a reference into the corresponding Slang code? I was not able to find the code location that lets a primitive fail.

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCase.java
```diff
@@ -43,11 +43,6 @@
     protected static SqueakImageContext image;
     protected static PointersObject nilClassBinding;
```
fniephaus: I don't know where `nilClassBinding` comes from, where is this field set?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecutePrimitiveContextNode.java
```diff
@@ -0,0 +1,74 @@
...
+                /* continue with fallback code. */
+            }
+        }
+        assert callPrimitiveNode.getSuccessorIndex() == CallPrimitiveNode.NUM_BYTECODES;
+        return doExecuteFresh(frame);
+    }
+
+    private CallPrimitiveNode getCallPrimitiveNode() {
+        if (callPrimitiveNode == null) {
+            callPrimitiveNode = (CallPrimitiveNode) fetchNextBytecodeNode(0);
```
fniephaus: `callPrimitiveNode` is not inserted correctly, this will cause some problems in the compiler.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ControlPrimitives.java
```diff
@@ -856,15 +856,19 @@ protected final Object doNamedPrimitiveWithArgs(final VirtualFrame frame, @Suppr
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 230)
     protected abstract static class PrimRelinquishProcessorNode extends AbstractPrimitiveNode implements BinaryPrimitive {
+        private static InterruptHandlerState istate;
```
fniephaus: Static state is to be avoided. In the future, we want to support running multiple images on the same GraalVM.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/util/InterruptHandlerState.java
```diff
@@ -203,4 +208,34 @@ public void reset() {
         pendingFinalizationSignals = false;
         semaphoresToSignal.clear();
     }
+
+    public void executeTrigger(final VirtualFrame frame, final SignalSemaphoreNode signalSemaporeNode) {
```
fniephaus: This `execute` method is no longer part of a node which is not a good idea. It's probably better to merge `InterruptHandlerState` into `InterruptHandlerNode` (other way around).

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ControlPrimitives.java
fmateoc: Well yes, but the current code is simply incorrect. It has guards that reference global mutable state (whether the external objects table is empty or not), and have nothing to do with the code object where they are used. As a result, asserts that the external objects is empty are generated, and the external objects table becomes non-empty dynamically at runtime, when an external semaphore is registered

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/util/InterruptHandlerState.java
fmateoc: I could have renamed the method, but the problem remains -see previous comment. Unfortunately we have to deal with global state here

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCase.java
fmateoc: It is set in AbstractSqueakTestCaseWithDummyImage, but it is declared in AbstractSqueakTestCase since it is used in the makeMethod... methods declared there.
All the makeMethod... methods are only called from subclasses of AbstractSqueakTestCaseWithDummyImage

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/HandleNonLocalReturnNode.java
fmateoc: Oh yes, many of the fixes are the result of running with an exception break on ClassCastException in Eclipse - this is a very useful Java debugging tip, since these are almost always programming errors (unlike say NullPointerException which are sometimes used for flow control)

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/ContextObject.java
fmateoc: Well, as the comment in #ensure: says:
"This method is *not*
	 implemented as a primitive.  Primitive 198 always fails.  The VM uses prim
	 198 in a context's method as the mark for an ensure:/ifCurtailed: activation."
Similar for 199, these are just markers, not implemented as primitives, so they can never be "in" the primitive call, so this avoids the call to getInstructionPointerForBytecodeLoop() which I assume is more expensive, for several methods which are used commonly for flow control.
It can probably be made even cheaper

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/RemoveProcessFromListNode.java
fmateoc: It is a problem because by leaving the first link not nil we are invalidating the check for an empty list (which relies on the first link being nil), which in turn is used in WakeHighestPriorityNode.
As a result, we end up trying to wake up the wrong process (from what should have been an empty list)

fniephaus: But why is this not a problem in the corresponding Slang code and OpenSmalltalkVM?

fmateoc: I have seen it happen. And isEmptyList is also used in SignalSemaphoreNode.
I don't know what to say about Slang, but it clearly does not respect the LinkedList protocol - it could be a bug

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ControlPrimitives.java
fniephaus: Could you please have a look at https://github.com/hpi-swa/graalsqueak/commit/902815fce0b4c4a961d054b6ddaf07caf5b3dd12? When I merged your recent change for `InterruptHandlerNode.java`, I didn't see the guard problem. It's well known and there are workarounds. Nonetheless, I think BranchProfiles are a better fit for this particular use case.

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
```diff
@@ -166,15 +166,12 @@ private static void resetSemaphoreLists() {
         resetList(NilObject.SINGLETON, interruptSema, "Interrupt semaphore");
         // The timer semaphore is taken care of in ensureTimerLoop, since the delays need to be
         // reset as well
+        final ArrayObject oldExternalObjects = (ArrayObject) image.getSpecialObject(SPECIAL_OBJECT.EXTERNAL_OBJECTS_ARRAY);
+        image.evaluate("[ ExternalObjectTable current " +
+                        "            initializeCaches;" +
+                        "            externalObjectsArray: (Smalltalk specialObjectsArray at: 39 put: (Array new: 20)) ] value");
```
fniephaus: Why not `ExternalObjectTable reset`?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ControlPrimitives.java
fmateoc: Thanks. I don't know enough about Truffle to comment about that part.
I do want to mention that we can also avoid entering the external semaphores signalling loop by doing a quick check if InterruptHandlerState's semaphoresToSignal is empty

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/ResumeProcessNode.java
fmateoc: InterpreterPrimitives>>primitiveResume does
	(objectMemory isContext: (objectMemory fetchPointer: SuspendedContextIndex ofObject: proc)) ifFalse:
		[^self primitiveFail].

fniephaus: Thanks! In that case, the check should be moved to `PrimResumeNode`...I'm on it.

fniephaus: Here it is: https://github.com/hpi-swa/graalsqueak/commit/47bdfd8055c1c254c2aa7a380ad0f2b6f505d508

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
```diff
@@ -77,6 +76,22 @@ private static void patchImageForTesting() {
...
+
+        final PointersObject priority40List = (PointersObject) ArrayObjectReadNode.getUncached().execute(lists, USER_PRIORITY_LIST_INDEX);
+        final Object firstLink = priority40List.instVarAt0Slow(LINKED_LIST.FIRST_LINK);
+        priority40List.instVarAtPut0Slow(LINKED_LIST.FIRST_LINK, newProcess);
+        if (firstLink == NilObject.SINGLETON) {
+            priority40List.instVarAtPut0Slow(LINKED_LIST.LAST_LINK, newProcess);
+        } else {
+            newProcess.instVarAtPut0Slow(PROCESS.NEXT_LINK, firstLink);
+        }
+        newProcess.instVarAtPut0Slow(PROCESS.LIST, priority40List);
```
fniephaus: Are you sure the tests run with the wrong priority? The following check suggests otherwise:

```
mx --dy /compiler squeak --code "Processor activeProcess priority" images/trunk64.image
[graalsqueak] Running Squeak/Smalltalk on OpenJDK 64-Bit GraalVM CE 19.2.1 (Graal-compiled)...
[graalsqueak] Image loaded in 1809ms.
Preparing image for headless execution...
[graalsqueak] Evaluating 'Processor activeProcess priority'...
[graalsqueak] Result: 40
```

fmateoc: they do run at the right priority initially. But after a timeout, which often happens during the sleep in relinquishProcess, the activeProcess is the idleProcess

fniephaus:
So I checked out your code and ran individual tests on my machine. Looks like you introduces some unexpected side-effect (similar to the guard/code generation issue for example). Considering the complexity of this PR, this will be very hard to debug. I've started to integrate some of your changes and so far, CI is still passing. Would you mind helping me split the rest of this PR into smaller changes that can be checked by CI individually? This way, I'm sure we're going to track down those unwanted side-effects.

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
fniephaus: Ah that's possible! I wonder if it would be easier to just reload the entire image/interpreter in this case? Trying to fix this in the JUnit test seems tedious and error-prone. Or maybe we should fail the build right after a timeout instead. What do you think?

fmateoc: Yes, I think failing the build right after a timeout might be the right solution.
Anyway tests are taking too long, and this will not let slip unexplained timeouts which almost always have something bad as a root cause

fmateoc:
> 
> 
> Which tests or other Smalltalk code is affected by this again? I'd like to leave a comment, so we don't forget.

The Decompiler ones, the Squeak stack trace looks like:

vvvvvvvvvvvvvvvvvv MessageNotUnderstood: UndefinedObject>>collect: vvvvvvvvvvvvvvvvvv
The time is 2019-11-11T16:28:59.423-06:00
UndefinedObject(Object)>>doesNotUnderstand: #collect:
StackInterpreter class>>primitiveTable
MethodNode>>printPrimitiveOn:
MethodNode>>printOn:
[] in MethodNode(Object)>>fullPrintString
String class(SequenceableCollection class)>>new:streamContents:
String class(SequenceableCollection class)>>streamContents:
MethodNode(Object)>>fullPrintString
MethodNode>>decompileString
DecompilerTests>>checkDecompileMethod:

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
fmateoc: because we also want to clean up the existing external semaphores registered in the previous (now dead) execution

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/process/RemoveProcessFromListNode.java
fmateoc: I think this should trigger the bug in Squeak:
| p | p := [Transcript show: 'bla1'] newProcess priority: 37. p resume. p suspend
But it doesn't. I assume the checks happening within the call to #isLiveContext: are sufficient to filter it out. Note that we do not do those extensive checks, we only guard against a nil suspendedContext

fmateoc:
and sure, I will go over the rest of the changes and try to submit them as smaller chunks

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
fmateoc: I mean the non-cached ones, inside the specialObjectsArray

fniephaus:
You closed the PR, so I'm assuming you no longer want me to look at your `dev` branch?

PR #83, opened 2019-10-31
fixed file scripts/template.graalsqueak.cmd

michelou:
Fixed quotes in handling of arguments/options.


fniephaus:
Thanks for the fix!

PR #88, opened 2019-11-12
Testing data setup fixes

fmateoc:
The bytecodes for primitive calls can only be at position 0 (1 in Smalltalk), rearrange the test data to avoid the error

Also the DecompilerTests end up using the StackInterpreter, which is loaded but not initialized in the current test image - this is a documentation-only fix though, it needs to be accompanied by the upload of a new test image with the initialization performed

The second commit deals with the scenario discussed about timeouts leaving both the state of the image in a bad state, and current worker threads still running. While this could be dealt with by stopping the entire suite from running, it is convenient for developers to be able to run the whole suite in one go and gather all failures. The code does reload the image in such cases (external timeouts), but otherwise it just does a more thorough cleanup of the state of the running image

### DIFF
--- a/src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
+++ b/src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
@@ -19,8 +19,10 @@ import org.junit.Assume;
 import org.junit.BeforeClass;
 
 import com.oracle.truffle.api.Truffle;
+import com.oracle.truffle.api.TruffleLogger;
 
 import de.hpi.swa.graal.squeak.exceptions.SqueakExceptions.SqueakException;
+import de.hpi.swa.graal.squeak.image.SqueakImageContext;
 import de.hpi.swa.graal.squeak.model.ArrayObject;
 import de.hpi.swa.graal.squeak.model.NativeObject;
 import de.hpi.swa.graal.squeak.model.NilObject;
@@ -28,16 +30,22 @@ import de.hpi.swa.graal.squeak.model.PointersObject;
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.LINKED_LIST;
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.PROCESS;
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.PROCESS_SCHEDULER;
+import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.SPECIAL_OBJECT;
 import de.hpi.swa.graal.squeak.nodes.ExecuteTopLevelContextNode;
 import de.hpi.swa.graal.squeak.nodes.accessing.ArrayObjectNodes.ArrayObjectReadNode;
+import de.hpi.swa.graal.squeak.shared.SqueakLanguageConfig;
 
 public class AbstractSqueakTestCaseWithImage extends AbstractSqueakTestCase {
+    private static final TruffleLogger LOG = TruffleLogger.getLogger(SqueakLanguageConfig.ID, SqueakImageContext.class);
     private static final int SQUEAK_TIMEOUT_SECONDS = 60 * 2;
-    private static final int TIMEOUT_SECONDS = SQUEAK_TIMEOUT_SECONDS + 2;
+    private static final int TIMEOUT_SECONDS = SQUEAK_TIMEOUT_SECONDS + 5;
     private static final int PRIORITY_10_LIST_INDEX = 9;
+    private static final int USER_PRIORITY_LIST_INDEX = 39;
     private static final String PASSED_VALUE = "passed";
 
     private static PointersObject idleProcess;
+    private static volatile boolean isClear;     // we have to be single-threaded, so the flag can
+                                                 // be static
 
     @BeforeClass
     public static void loadTestImage() {
@@ -45,11 +53,13 @@ public class AbstractSqueakTestCaseWithImage extends AbstractSqueakTestCase {
         loadImageContext(imagePath);
         image.getOutput().println("Test image loaded from " + imagePath + "...");
         patchImageForTesting();
+        isClear = true;
     }
 
     @AfterClass
     public static void cleanUp() {
         idleProcess = null;
+        image.interrupt.reset();
         destroyImageContext();
     }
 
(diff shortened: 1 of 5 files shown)


src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/AbstractSqueakTestCaseWithImage.java
```diff
@@ -19,37 +19,47 @@
...
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.LINKED_LIST;
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.PROCESS;
 import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.PROCESS_SCHEDULER;
+import de.hpi.swa.graal.squeak.model.layout.ObjectLayouts.SPECIAL_OBJECT;
 import de.hpi.swa.graal.squeak.nodes.ExecuteTopLevelContextNode;
 import de.hpi.swa.graal.squeak.nodes.accessing.ArrayObjectNodes.ArrayObjectReadNode;
+import de.hpi.swa.graal.squeak.shared.SqueakLanguageConfig;
 
 public class AbstractSqueakTestCaseWithImage extends AbstractSqueakTestCase {
+    private static final TruffleLogger LOG = TruffleLogger.getLogger(SqueakLanguageConfig.ID, SqueakImageContext.class);
```
fniephaus: The logger should usually be attached to the class it is located in (AbstractSqueakTestCaseWithImage in this case).

fniephaus:
I have cherry-picked https://github.com/hpi-swa/graalsqueak/pull/88/commits/aa80c3b68590856fb9fc23919651e591fd7516e7 and updated the test image accordingly. Please note that `slowly_(passing|failing)` tests are skipped by our CI. So if you run tests locally, you may see failures as we don't update slow tests as often.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/ContextObject.java
```diff
@@ -324,7 +325,11 @@ public AbstractSqueakObject getSender() {
         final Object value = FrameAccess.getSender(truffleFrame);
         if (value instanceof FrameMarker) {
             getBlockOrMethod().getDoesNotNeedSenderAssumption().invalidate("Sender requested");
-            return ((FrameMarker) value).getMaterializedContext();
+            try {
+                return ((FrameMarker) value).getMaterializedContext();
+            } catch (final SqueakException e) {
+                return NilObject.SINGLETON;
+            }
```
fniephaus: I don't think it's a good idea to catch `SqueakException` here, it's probably better to move this logic into `getMaterializedContext()`.

fmateoc:
I have some other additional logging, which for me was extremely helpful in debugging, but they are not fixes, and I assume they are not zero-overhead. Should I submit them?

fniephaus:
> I have some other additional logging, which for me was extremely helpful in debugging, but they are not fixes, and I assume they are not zero-overhead. Should I submit them?

Logging is only zero-overhead when removed by the Graal compiler, so interpreter performance may be impacted, especially when logging in performance-critical locations (such as scheduling). Also, logging sometimes feels very specific to a problem one wants to solve and does not help much when trying to take some other (related) problem. But we could move your logging infrastructure to a branch, so the work is persisted and we can look it up in the future. What do you think? If you create a dedicated branch, I could push it, for example, to `fmateoc/logging`.

I'm still think about your changes to `AbstractSqueakTestCaseWithImage.java`. It's quite a lot for restoring sanity after a failure and we can never be sure that it works. Maybe it'd be better to just fail hard as soon as we get a timeout. That would certainly make things a bit easier in our testing infrastructure. What do you think?

fmateoc:
Yes, I think a separate branch for logging makes sense. From my perspective the changes to AbstractSqueakTestCaseWithImage are in a similar situation - they were invaluable for my testing and debugging. But I am not sure if I should put them together with logging. 
I agree that for the testing infrastructure a fail hard when getting a TimeoutException is the best solution

fmateoc:
Also there are two separate issues that AbstractSqueakTestCaseWithImage deals with. One is the timeout/multithreading issue, the other is the state of the reused image. If we want to keep reusing the image for multiple test, and since we already attempt to clean up the processes, I think we should also clean up semaphores

fmateoc:
I am also not talking about a theoretical issue. Right now, even without timeouts, there are tests that fail within the suite, but succeed when run in isolation. I think that's not good

fniephaus:
> Yes, I think a separate branch for logging makes sense. From my perspective the changes to AbstractSqueakTestCaseWithImage are in a similar situation - they were invaluable for my testing and debugging. But I am not sure if I should put them together with logging.
I agree that for the testing infrastructure a fail hard when getting a TimeoutException is the best solution

You can split your changes and we could add multiple branches. Also, the smaller your commits are, the easier it is to cherry pick them.

> Also there are two separate issues that AbstractSqueakTestCaseWithImage deals with. One is the timeout/multithreading issue, the other is the state of the reused image. If we want to keep reusing the image for multiple test, and since we already attempt to clean up the processes, I think we should also clean up semaphores

Again, the problem is that you addressed these two separate issues in the same commit. It'd be great if you could split the commit into two: one for cleaning up semaphores and another one for failing in case of timeouts. You don't have to, but I also don't mind multiple PRs in the future. It's just easier than reviewing a growing PR. Apart from that, we shouldn't forget about your ideas for refactoring the bytecode loop in terms of primitive calls. :)

Thanks again for your contributions! Your help is greatly appreciated.

fniephaus:
> I am also not talking about a theoretical issue. Right now, even without timeouts, there are tests that fail within the suite, but succeed when run in isolation. I think that's not good

I am assuming you're always talking about Windows, because I haven't seen such problems on Linux/macOS. As discussed before, we should set up a Windows CI job, so we can identify and reproduce these kind of issues.

fmateoc:
Thank you for your patience and for your prompt and thoughtful replies. I agree with everything you said and I will try to improve my submissions in the future.
Now, just to recap the growing to-do list:
1. a logging branch
2. separate concerns in AbstractSqueakTestCaseWithImage 
3. make timeouts a hard stop in CI testing
4. setting up a Windows CI
5. new PR with testcase and fix for process removal from lists
6. new PR with testcases and fix for cannotReturn (it turns out the handling should not happen in ResumeContextNode but in NonLocalReturn, but it's still not working properly yet - right now it's not working but it never gets invoked, so you might leave it there for now, up to you)
7. new PR for refactoring the bytecode loop in terms of primitive calls

I am happy to slowly (I do have a day job and I will be out of town this coming weekend) work on all of them, and I am happy to learn more in order to take care of 3 and 4, but I don't have a lot of experience with them, so it might take some time. Anyway, if you think these items need to be prioritized (also feel free to add to them, I might have missed something), please let me know

fniephaus:
I've opened a tracking issue (https://github.com/hpi-swa/graalsqueak/issues/89) for your to-do list. There's no rush at all, so please take your time. :) I might try to set up the Windows CI build job in the next few days, which hopefully will help us to better understand the issues you have identified.

fmateoc:
Excellent, thank you!

PR #91, opened 2019-11-17
this fixes the cannotReturn implementation and adds a couple of tests for it.

fmateoc:
For the testcases I also had to fix the doItContext creation - while a doItMethod does not have arguments or copiedValues, it may have temporaries. Also the class is UndefinedObject, but the receiver is nil

### DIFF
--- a/src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/DispatchEagerlyNode.java
+++ b/src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/DispatchEagerlyNode.java
@@ -25,6 +25,7 @@ import de.hpi.swa.graal.squeak.util.FrameAccess;
 @NodeInfo(cost = NodeCost.NONE)
 public abstract class DispatchEagerlyNode extends AbstractNodeWithCode {
     protected static final int INLINE_CACHE_SIZE = 6;
+    protected static final boolean TRUE = true;
 
     protected DispatchEagerlyNode(final CompiledCodeObject code) {
         super(code);
@@ -57,7 +58,7 @@ public abstract class DispatchEagerlyNode extends AbstractNodeWithCode {
                     limit = "INLINE_CACHE_SIZE", assumptions = {"cachedMethod.getCallTargetStable()"}, replaces = {"doPrimitiveEagerly"})
     protected static final Object doDirectWithSender(final VirtualFrame frame, @SuppressWarnings("unused") final CompiledMethodObject method, final Object[] receiverAndArguments,
                     @SuppressWarnings("unused") @Cached("method") final CompiledMethodObject cachedMethod,
-                    @Cached("create(code)") final GetOrCreateContextNode getOrCreateContextNode,
+                    @Cached("create(code, TRUE)") final GetOrCreateContextNode getOrCreateContextNode,
                     @Cached("create(cachedMethod.getCallTarget())") final DirectCallNode callNode) {
         return callDirect(callNode, cachedMethod, getOrCreateContextNode.executeGet(frame), receiverAndArguments);
     }
@@ -70,7 +71,7 @@ public abstract class DispatchEagerlyNode extends AbstractNodeWithCode {
 
     @Specialization(guards = "!method.getDoesNotNeedSenderAssumption().isValid()", replaces = {"doDirect", "doDirectWithSender"})
     protected static final Object doIndirectWithSender(final VirtualFrame frame, final CompiledMethodObject method, final Object[] receiverAndArguments,
-                    @Cached("create(code)") final GetOrCreateContextNode getOrCreateContextNode,
+                    @Cached("create(code, TRUE)") final GetOrCreateContextNode getOrCreateContextNode,
                     @Cached final IndirectCallNode callNode) {
         return callIndirect(callNode, method, getOrCreateContextNode.executeGet(frame), receiverAndArguments);
     }
--- a/src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
+++ b/src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
@@ -31,7 +31,9 @@ import de.hpi.swa.graal.squeak.exceptions.ProcessSwitch;
 import de.hpi.swa.graal.squeak.exceptions.Returns.NonLocalReturn;
 import de.hpi.swa.graal.squeak.exceptions.Returns.NonVirtualReturn;
 import de.hpi.swa.graal.squeak.model.CompiledCodeObject;
+import de.hpi.swa.graal.squeak.model.ContextObject;
 import de.hpi.swa.graal.squeak.nodes.ExecuteContextNodeFactory.TriggerInterruptHandlerNodeGen;
+import de.hpi.swa.graal.squeak.nodes.accessing.AbstractPointersObjectNodes.AbstractPointersObjectReadNode;
 import de.hpi.swa.graal.squeak.nodes.bytecodes.AbstractBytecodeNode;
 import de.hpi.swa.graal.squeak.nodes.bytecodes.JumpBytecodes.ConditionalJumpNode;
 import de.hpi.swa.graal.squeak.nodes.bytecodes.JumpBytecodes.UnconditionalJumpNode;
@@ -41,7 +43,6 @@ import de.hpi.swa.graal.squeak.nodes.bytecodes.ReturnBytecodes.AbstractReturnNod
 import de.hpi.swa.graal.squeak.nodes.bytecodes.SendBytecodes.AbstractSendNode;
 import de.hpi.swa.graal.squeak.nodes.context.frame.FrameStackInitializationNode;
 import de.hpi.swa.graal.squeak.shared.SqueakLanguageConfig;
-import de.hpi.swa.graal.squeak.util.ArrayUtils;
 import de.hpi.swa.graal.squeak.util.FrameAccess;
 import de.hpi.swa.graal.squeak.util.InterruptHandlerNode;
 import de.hpi.swa.graal.squeak.util.SqueakBytecodeDecoder;
@@ -93,7 +94,9 @@ public class ExecuteContextNode extends AbstractNodeWithCode implements Instrume
         final boolean enableStackDepthProtection = enableStackDepthProtection();
         try {
             if (enableStackDepthProtection && code.image.stackDepth++ > STACK_DEPTH_LIMIT) {
-                throw ProcessSwitch.createWithBoundary(getGetOrCreateContextNode().executeGet(frame));
+                final ContextObject context = getGetOrCreateContextNode().executeGet(frame);
+                context.setProcess(code.image.getActiveProcess(AbstractPointersObjectReadNode.getUncached()));
+                throw ProcessSwitch.createWithBoundary(context);
             }
             frameInitializationNode.executeInitialize(frame);
             return startBytecode(frame);
@@ -141,7 +144,7 @@ public class ExecuteContextNode extends AbstractNodeWithCode implements Instrume
     private GetOrCreateContextNode getGetOrCreateContextNode() {
         if (getOrCreateContextNode == null) {
             CompilerDirectives.transferToInterpreterAndInvalidate();
-            getOrCreateContextNode = insert(GetOrCreateContextNode.create(code));
+            getOrCreateContextNode = insert(GetOrCreateContextNode.create(code, false));
         }
         return getOrCreateContextNode;
     }
@@ -165,8 +168,7 @@ public class ExecuteContextNode extends AbstractNodeWithCode implements Instrume
                         return callPrimitiveNode.primitiveNode.executePrimitive(frame);
                     } catch (final PrimitiveFailed e) {
                         getHandlePrimitiveFailedNode().executeHandle(frame, e.getReasonCode());
-                        LOG.log(Level.FINE, () -> callPrimitiveNode.primitiveNode +
-                                        " (" + ArrayUtils.toJoinedString(", ", FrameAccess.getReceiverAndArguments(frame)) + ")");
+                        LOG.log(Level.FINE, () -> "Failed primitive " + callPrimitiveNode.primitiveNode.getClass().getName());
                         /* continue with fallback code. */
                     }
                 }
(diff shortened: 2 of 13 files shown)


src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
```diff
@@ -69,7 +76,37 @@ protected final Object doNonLocalReturn(final VirtualFrame frame) {
         @Specialization(guards = {"isCompiledBlockObject(code)"})
         protected final Object doClosureReturn(final VirtualFrame frame) {
             // Target is sender of closure's home context.
-            throw new NonLocalReturn(getReturnValue(frame), FrameAccess.getClosure(frame).getHomeContext().getFrameSender());
+            final ContextObject homeContext = FrameAccess.getClosure(frame).getHomeContext();
+            final ContextObject[] firstUsableContext = new ContextObject[1];
+            final boolean homeContextNotOnTheStack = null == Truffle.getRuntime().iterateFrames(frameInstance -> {
```
fniephaus: Thanks for the PR! Just briefly looked at this and saw that you use `iterateFrames()` for all non-local returns. That's probably going to be *very* slow. Instead, we may want to add a flag to `ContextObject` that indicates if the context is active on the stack or not. We've been doing something similar in RSqueak (see [here](https://github.com/hpi-swa/RSqueak/blob/7d9b6649efeb1d1fff9f2ecf4084d3dd18082022/rsqueakvm/storage_contexts.py#L16-L32)).

fmateoc: I had a bug that I had to fix, so I re-pushed.
Now it passes the CI tests (it does show a Checkstyle error, but it's not something I modified or introduced)
I also added some specializations to try to avoid as much as possible materialization and the frame iteration.
I am afraid a flag on context would not help. What we are looking for with this iteration, and which is the "common" case for a cannotReturn situation, is a context that is valid and active on a stack, just not the right one (it's in a different process, see the testcases)

fmateoc: What I think would be interesting to pursue, and I am sure that would help performance in general, not just by reducing these iterations, is targeted and (more?) aggressive inlining of the methods around closures, until the closures disappear - then the non-local returns disappear as well.

fniephaus: > But I am afraid that a flag on the context would not help. What we are looking for within this iteration, and which is the "common" case for a cannotReturn, is a context that is valid and active on a stack, just not the right one (e.g. it's in a different process, see the testcases)

By a flag on the context that indicates whether it is active or not, I meant active on the Truffle stack. Only one process can be active on the Truffle stack, so this should work, right?

> What I think would be interesting to pursue, and I am sure that would help performance in general, not just by reducing these iterations, is targeted and (more?) aggressive inlining of the methods around closures, until the closures disappear - then the non-local returns disappear as well.

The Graal compiler inlines closures already. However, I am not sure how well it can figure out the stack walking magic (`iterateFrames`). AFAIK, `iterateFrames` should really only be used the first time on the slow path, and afterwards, there should be an alternative version that can be compiled as part of the fast path.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
```diff
@@ -27,6 +35,21 @@ protected AbstractReturnNode(final CompiledCodeObject code, final int index) {
             super(code, index);
         }
 
+        protected final static AbstractSqueakObject findOnSenderChain(final ContextObject from, final ContextObject target) {
```
fniephaus: This needs to be `static final` for the checkstyle test to pass.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
fmateoc: You are probably right, the flag should work, at least for "normal" code. I am not sure how e.g. copyStack-based frameworks like Seaside are (or at least used to be) implemented - anyway, such code would probably be a good testcase.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
fmateoc: Thank you, done

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
fniephaus: Implementing the flag should be relatively straightforward (compared to other approaches), shall we give it a try and take it from there? BTW: we ran the Seaside test suite at ESUG and were surprised that only a couple of Context-related tests are not passing at the moment. Not sure whether you've seen [this](https://github.com/marschall/seaside-servlet-squeak). :)

fmateoc: Very nice, I had not seen it, thanks!
I'll try to implement the flag

fmateoc:
Not sure if you had something like this in mind.
I ended up adding not a flag to contexts but their associated process (the process with the (potentially conceptual for active process) stack trace they live on). I think this has similar runtime performance impact, but greater potential for other uses in the future

fniephaus:
> Not sure if you had something like this in mind.
I ended up adding not a flag to contexts but their associated process (the process with the (potentially conceptual for active process) stack trace they live on). I think this has similar runtime performance impact, but greater potential for other uses in the future

Cool that you got this working without `iterateFrames`! My only concern, however, is that you are using a `Map` in `SqueakImageContext`. This could cause two different problems: 1) it may introduce a leak (you never know what kind of code is running, not sure we can ensure there won't be any leaks). And 2) `Map` is highly recursive which the partial evaluator does not like much. So there could again be a performance problem. Having learned from my own premature optimizations, I wonder what other uses you have in mind for this. AFAICT the OpenSmalltalkVM does not have such an infrastructure, but maybe you have a good reason we want this in GraalSqueak. With a flag, on the other hand, we wouldn't hit the problems I mentioned before.

fmateoc:
The map is not part of the infrastructure, it is only used during image load time, as just another step in initializeAfterLoadingImage(), before any activity starts, and it is cleared right there as well

fmateoc:
It is used just as a temporary holder for the suspended contexts when filling the process objects, since at that time contexts are not yet initialized.

fmateoc:
The only thing that is added as part of the infrastructure is a process field (instead of a flag) on the context

fniephaus:
> The map is not part of the infrastructure, it is only used during image load time, as just another step in initializeAfterLoadingImage(), before any activity starts, and it is cleared right there as well.
It is used just as a temporary holder for the suspended contexts when filling the process objects, since at that time contexts are not yet initialized.

Sorry my bad...you are right!

> The only thing that is added as part of the infrastructure is a process field (instead of a flag) on the context

What other uses of this new reference do you have in mind?

fmateoc:
for one, I think this could also be used for the other places where we do iterate over frames - finding the unwind and exception handling blocks

fniephaus:
Hi @fmateoc,
Sorry for the delay...I haven't forgotten about this PR. I've just been very busy with a couple of other things. Plus our benchmark infrastructure is currently broken due to the upgrade to `19.3`. But once that's fixed, I'll check if this PR introduces any significant performance penalties and will try to integrate it.

fmateoc:
No problem, I knew when I started that this was an open-source project, with everyone a volunteer, plus with apparently only you as an active developer, and you might have other priorities as well. Anyway, I am interested enough in the project that I did not just stop and wait for your feedback - I continued on my own, I think I am making good progress, and, quite frankly, it feels liberating to not be subject to the restrictions of having to have everything approved by somebody else.

fniephaus:
Well that's fair enough. I'm going to re-open this PR as it adds support for cannotReturn. Handling error cases correctly is important, but as you suspected, I have other priorities at the moment (in no particular order):
- Supporting language interop
- Improving performance
- Staying up-to-date with GraalVM releases
- Adding missing primitives (e.g snapshot), plugins, and other core features

As I mentioned before, I'm particularly cautious and hesitant when it comes to changes to the interpreter loop or bytecode implementations as these are performance-critical parts in GraalSqueak.

Nonetheless, I'd also like to understand your goals and priorities, and I would really appreciate any contributions from you. So if we could work something out that works for both of us, that'd be great. Maybe we could talk about this over a video call?

fmateoc:
Sure, I'd be happy to talk, but I am right now on vacation in a remote national park (Big Bend, Texas), with very poor connectivity. I'll be back in NY on Jan 4th.
Anyway, my priorities regarding this project are (I think for me they do have a ranking)
1. adhering strictly to the language semantics
2. improving performance 
3. improving runtime debuggability 
4. improving source readability
5. staying up-to-date with graalvm
6. adding missing primitives - indeed, I think snapshot would be an important one

fniephaus:
@fmateoc Sounds good, looks like our goals are not that different. With regard to 1. and 2., however, we had to make some trade-offs (e.g. interrupt handling).
Nonetheless, enjoy your days off! Would you have some time on Jan 7th for a call? We can also move this sort of conversation to a private chat in the GraalVM Slack.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/DispatchEagerlyNode.java
```diff
@@ -25,6 +25,7 @@
 @NodeInfo(cost = NodeCost.NONE)
 public abstract class DispatchEagerlyNode extends AbstractNodeWithCode {
     protected static final int INLINE_CACHE_SIZE = 6;
+    protected static final boolean TRUE = true;
```
fniephaus: This is no longer necessary since [Truffle `19.3`](https://github.com/oracle/graal/blob/094363c11f3cf4882e0f90fc1788d19ded9f703d/truffle/CHANGELOG.md#version-1930). I'll take care of it :)

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -165,8 +168,7 @@ private Object startBytecode(final VirtualFrame frame) {
                         return callPrimitiveNode.primitiveNode.executePrimitive(frame);
                     } catch (final PrimitiveFailed e) {
                         getHandlePrimitiveFailedNode().executeHandle(frame, e.getReasonCode());
-                        LOG.log(Level.FINE, () -> callPrimitiveNode.primitiveNode +
-                                        " (" + ArrayUtils.toJoinedString(", ", FrameAccess.getReceiverAndArguments(frame)) + ")");
+                        LOG.log(Level.FINE, () -> "Failed primitive " + callPrimitiveNode.primitiveNode.getClass().getName());
```
fniephaus: I'm assuming you changed this because you saw some multiline string arguments being printed in your console?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
```diff
@@ -141,7 +144,7 @@ public final Object executeResumeInMiddle(final VirtualFrame frame, final long i
     private GetOrCreateContextNode getGetOrCreateContextNode() {
         if (getOrCreateContextNode == null) {
             CompilerDirectives.transferToInterpreterAndInvalidate();
-            getOrCreateContextNode = insert(GetOrCreateContextNode.create(code));
+            getOrCreateContextNode = insert(GetOrCreateContextNode.create(code, false));
```
fniephaus: This seems to be the only time, `fromActiveProcess` is set to `false`. Is this correct? If yes, why did you have to set it to `false` here? The node is used to trigger a `ProcessSwitch` to itself. This way, we protect GraalSqueak from a `StackOverflowError`.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/ReturnBytecodes.java
```diff
@@ -67,9 +75,33 @@ protected final Object doNonLocalReturn(final VirtualFrame frame) {
...
-        protected final Object doClosureReturn(final VirtualFrame frame) {
+        protected final Object doClosureReturnFromMaterialized(final VirtualFrame frame) {
             // Target is sender of closure's home context.
-            throw new NonLocalReturn(getReturnValue(frame), FrameAccess.getClosure(frame).getHomeContext().getFrameSender());
+            final ContextObject homeContext = FrameAccess.getClosure(frame).getHomeContext();
+            assert homeContext.getProcess() != null;
+            final boolean homeContextNotOnTheStack = homeContext.getProcess() != code.image.getActiveProcess(readNode);
+            final Object caller = homeContext.getFrameSender();
+            if (caller == NilObject.SINGLETON || homeContextNotOnTheStack) {
+                getCannotReturnNode().executeSend(frame, getGetOrCreateContextNode().executeGet(frame), getReturnValue(frame));
```
fniephaus: First, I  thought this needs a `BranchProfile`, but `getCannotReturnNode()` basically acts like a `BranchProfile` here.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
fmateoc: I also saw Truffle protesting against accessing the frame (throwing exceptions) - I thought it was not worth it

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/ExecuteContextNode.java
fmateoc: On the contrary, the stack overflow seems to be the rare exception. The node is actually used in the process switch branch as well as the non virtual return, precisely the places where we switch process to some other process. I thought the rarely used stack overflow check was not worth treating specially

fmateoc:
January 7th sounds good. GraalVM slack also sounds good :)

fniephaus:
Looks like the [AWFY benchmarks](https://github.com/smarr/are-we-fast-yet/) show no performance regressions (probably because they are not using non-local returns). Also, I've cleaned a few things up (https://github.com/hpi-swa/graalsqueak/commit/bf7ba33ea29d5721e87dbe40f4eb668d1b8900bf) and moved the `suspendedContexts` map to the image reader (https://github.com/hpi-swa/graalsqueak/commit/e633184319f23dd8b5a163cabc8ef3ead3ecba7d). If all tests are still passing, it's time to get this merged! :)

PR #97, opened 2020-01-10
Cleaned up changes from @fmateoc

fniephaus:


src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/CompiledMethodObject.java
```diff
@@ -139,22 +139,27 @@ public boolean hasMethodClassSlow() {
...
     }
 
     public ClassObject getMethodClassSlow() {
         CompilerAsserts.neverPartOfCompilation();
-        return getMethodClass(AbstractPointersObjectReadNode.getUncached());
+        final AbstractPointersObjectReadNode readNode = AbstractPointersObjectReadNode.getUncached();
+        if (hasMethodClass(readNode)) {
+            return getMethodClass(readNode);
+        }
+        return null;
```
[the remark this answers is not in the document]
fmateoc: since we want to be able to load random Squeaq images, I think it's better to be tolerant :)

fmateoc:
Well, the tests succeed now, but the image build fails. :(

fniephaus:
Great! I guess you're talking about native-image builds, it's quite ambiguous to talk about images in the context of GraalSqueak. 😅

Anyway, it's easy to understand these kind of problems and to fix them. The error ["18609 method(s) included for runtime compilation"](https://travis-ci.org/hpi-swa/graalsqueak/jobs/635587230#L17093) means that code size exploded during AOT compilation. This usually happens when something highly-recursive (e.g. string or hash operations) is part of partial evaluation (PE), which shouldn't be part of PE.  [The stack trace](https://travis-ci.org/hpi-swa/graalsqueak/jobs/635587230#L17141) usually reveals where the explosion happened. I've already integrated the commit that is causing the explosion in your case. But I've made sure the `@TruffleBoundary` is still in place (see https://github.com/hpi-swa/graalsqueak/commit/b2aa56c682944057a612fb9fc6e89f8d63f9daee). And that's what these boundaries are for: excluding code from PE. I hope this makes sense, the fix is quite easy!

fmateoc:
Sure, thanks. I had already encountered an explosion caused by HashMap before, and I googled it and found out about the recursion problem, but it was only once, and it was reported at runtime - I didn't make the connection. Now that the lesson has been reinforced, I hopefully won't forget about it again :)

fniephaus:
It's relatively easy to forget about these boundaries, which is one reason we build native-images on our CI infrastructure. :)

src/de.hpi.swa.graal.squeak.test/src/de/hpi/swa/graal/squeak/test/tests.properties
```diff
@@ -699,7 +699,7 @@ ColorTest>>testMultiplyByNumber=passing
 ColorTest>>testNew=passing
 ColorTest>>testPrintHtmlString=passing
 ColorTest>>testUnCategorizedMethods=passing
-CompiledMethodComparisonTest>>testHash=passing
+CompiledMethodComparisonTest>>testHash=not_terminating
```
fniephaus: I will add the info as a comment to the properties file, so the next time we look at this, we know why it's marked as `not_terminating`.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java
```diff
@@ -66,6 +66,6 @@ protected final boolean isDoubleOrIllegal(final Frame frame) {
     }
 
     private static boolean verifyWrite(final Object value) {
-        return value != null && !(value instanceof Byte || value instanceof Integer || value instanceof Float);
```
fniephaus: This was used to detect illegal writes of Java primitives not used in GraalSqueak to avoid that they propagate through the system. I'll turn it into an assertion.

fniephaus: It's already an assertion, I hope I have clarified this better in the code with https://github.com/hpi-swa/graalsqueak/commit/3231e4b1c33418c1124e7451db87567b9cf78ecc.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java
```diff
@@ -2542,10 +2542,14 @@ public Object primitiveCopyBits(final PointersObject bbObj, final long factor) {
...
         }
-        copyBits(factor);
-        assert !failed();
-        showDisplayBits();
-        assert !failed();
+        try {
+            copyBits(factor);
+            assert !failed();
+            showDisplayBits();
+            assert !failed();
```
fniephaus: Interesting find! Which tests caused the assertions errors? Would it be enough to turn each of these two `assert`s into this?
```java
if (failed()) {
  PrimitiveFailed.andTransferToInterpreter();
}

Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java]: but there are other asserts too, embedded in other methods, this is just the top one. Actually the asserts hit and raised were the ones that I ended up removing (in copyLoop())

File: src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java (hunk 3 above)
Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java]: I am afraid this does not work for integer types, it only works for null and Boolean and presumably Float.
There are three tests that are failing because of this, and they are legitimate: for a method with an argument typed as Object, the argument comes as the value 40. Who knows where and who did the boxing, and how can the test for Byte and Short and Integer fail for it? Mind you, the value is also red from a slot, which also legitimately is the kind Object (everything on the stack is). So how does it gets boxed when read as on Object? Certainly not as a Long

Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java]: The tests are regressions from Squeak, and they are the PNGReadWriterTest failing ones

Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java]: Note that, at the very least, we also use the Int slot kind explicitly, for pc and stackp, and those values also legitimately get onto the stack

Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java]: Well, I saw it triggered by some testcases that worked after removing the check. But I confess I did not understand what was happening. And I did have a harrowing debugging experience when in my LargeInteger changes I returned a 0 instead of a 0L :)

File: src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java (hunk 4 above)
Comment (fniephaus) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java]: If we want to be correct, we'd need to do the `PrimitiveFailed.andTransferToInterpreter();` dance for every assertion. I've decided to use assertions to be able to debug them, because sometimes (e.g. during startup) BitBlt is supposed to fail. Let's leave all assertions in place unless they are the reasons some tests are failing or some images are crashing. Then, we should debug them or let the primitive fail.

File: src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java (hunk 3 above)
Comment (fniephaus) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/context/frame/FrameSlotWriteNode.java]: Yes, I'm still making this mistake sometimes. Integers are only used in very rare places and should never "escape" via the stack.

File: src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java (hunk 4 above)
Comment (fmateoc) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java]: but doesn't my solution achieve that as well? The tests should fail as well with a primitive failure, and if you run under debug with trapping AssertionError, you catch it when it's thrown. An assert is equivalent with a vm crash, i don't think it's better than a primitive failure - this errors do happen on the vm side, inside the primitive implementation. That's what primitive failures are for

Comment (fniephaus) [reply src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/plugins/BitBlt.java]: Right, I forgot that BitBlt and Balloon are behind Truffle boundaries and therefore excluded from Truffle compilation. Shall we log the assertion error to stdout, so that users will see and can report them?

File: src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/CompiledCodeObject.java [hunk 5]
```diff
@@ -215,6 +215,23 @@ public final FrameSlot getStackSlot(final int i) {
...
+                stackSlots[i] = frameDescriptor.addFrameSlot(i + 1, FrameSlotKind.Long);
+            } else if (value instanceof Double) {
+                stackSlots[i] = frameDescriptor.addFrameSlot(i + 1, FrameSlotKind.Double);
+            } else {
+                stackSlots[i] = frameDescriptor.addFrameSlot(i + 1, FrameSlotKind.Object);
+            }
+        }
+        return stackSlots[i];
+    }
+
```
fniephaus: I'm afraid we won't gain much performance from this as adding stack slots is a slow path operation anyway, so it doesn't really matter if we set the type later in the same AST or right after creation. The stack slots logic is already quite complex, so I rather not add additional logic we have to keep in tact.

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/bytecodes/StoreBytecodes.java
```diff
@@ -22,18 +22,18 @@
 public final class StoreBytecodes {
 
     private abstract static class AbstractStoreIntoAssociationNode extends AbstractStoreIntoNode {
-        protected final long variableIndex;
+        protected final Object literal;
```
fniephaus: Literals of a method can potentially change, so I don't think it's a good idea to cache them here.

fniephaus:
The commits on this branch have been merged into `fmateoc:merge/large-integers`.

PR #99, opened 2020-01-15
large integers fixes and optimizations

fmateoc:
this adds bitLength to LargeIntegerObject to allow for cheap checks and optimizations. It then uses it to avoid LargeIntegerObject allocations when they can fit in longs
it also fixes a couple of bugs in comparisons, as well as a bug in shift, which is not a shift magnitude for negative numbers


fmateoc:
I don't have those errors locally.
While trying to check for differences between the branches, I noticed that in your commit 1c42ca0d9b50370b8a87fd30a0c72e2fd1888bcd (further layout fixes: canStore should work for nil, unset should happen whenever oldLocation is discarded), you only took a part of it, you did not take the SlotLocation changes. I think those are important

fmateoc:
Now that's really weird. As I am still on graalvm 19.2.1, I could not run directly the code on the branch from where I made the pull request. But now, seeing that the errors still occur, I checked out the exact same code and just made the very minimal changes to compile and run on 19.2.1 - changed getPublicTruffleFile() calls to getTruffleFile() calls and created static variables instead of true inside guards. Otherwise the code is exactly the same and the first testcase, DoubleByteArrayTest>>testCannotPutTooLargeValue, which fails with a non-timeout assert in CI, succeeds locally (on 19.2 though)

fmateoc:
I also tried to move to 19.3.1, but apparently the way to invoke have changed, I cannot run the tests with my existing Eclipse launch configurations

fmateoc:
Although CI is failing, I am now happy with the results. Other than a few tests that time out - I don't know what that means, but I doubt it's related to the code changes - there are a couple of idiotic tests in LargeNegativeIntegerTest that are "testing" invalid 0-valued un-normalized large integers. Such a number breaks so many assumptions: it responds false to #isZero, they are also of course not equal to 0, plus you have now multiple "integers" in the image with the same 0 value. I will remove those two tests as part of this PR

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ArithmeticPrimitives.java
```diff
@@ -46,12 +47,7 @@ public PrimAddNode(final CompiledMethodObject method) {
...
 
-        @Specialization(rewriteOn = ArithmeticException.class)
-        protected static final long doLong(final long lhs, final long rhs) {
-            return Math.addExact(lhs, rhs);
-        }
-
-        @Specialization(replaces = "doLong")
+        @Specialization
         protected final Object doLongWithOverflow(final long lhs, final long rhs) {
             return LargeIntegerObject.add(method.image, lhs, rhs);
```
fniephaus: Now, that all specializations failing with an `ArithmeticException` are gone, we must make sure that operations such as `add` are not behind a Truffle boundary. Otherwise, we have a performance problem.

fniephaus: I think it makes more sense to keep the original specializations that can throw a `ArithmeticException`. As soon as such an exception is thrown, we fall back to the large integer implementation, which retries the long operation, but behind a boundary. The next time, it will go straight to the large integer operation. I think that's good!

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/nodes/primitives/impl/ControlPrimitives.java
```diff
@@ -583,16 +588,76 @@ protected final long doGC(@SuppressWarnings("unused") final Object receiver) {
...
+                    final long accumulatedCollectionTime = bean.getCollectionTime();
+                    if (accumulatedCollectionTime != -1) {
+                        enabledBeansTimes.put(bean, accumulatedCollectionTime);
+                    }
+                }
+            }
+            long elapsed = 0;
+            final long start = System.nanoTime();
+            try {
+                dcmd.invoke("gcClassHistogram", new Object[]{new String[]{}}, new String[]{"[Ljava.lang.String;"});
```
fniephaus: Could we use `gcRun` instead?

src/de.hpi.swa.graal.squeak/src/de/hpi/swa/graal/squeak/model/ContextObject.java
```diff
@@ -446,7 +446,7 @@ public void setReceiver(final Object value) {
 
     @TruffleBoundary
     public Object atTemp(final int index) {
-        return NilObject.nullToNil(truffleFrame.getValue(getBlockOrMethod().getStackSlot(index)));
+        return NilObject.nullToNil(FrameUtil.getObjectSafe(truffleFrame, getBlockOrMethod().getStackSlot(index)));
```
fniephaus: Well here, you have to use `truffleFrame.getValue()` I believe because you didn't check `truffleFrame.isObject(slot)`.

fniephaus:
Large integer fixes and optimizations are cherry-picked via https://github.com/hpi-swa/graalsqueak/commit/aaae7cf0d4627053875a4954ef31e18d06b8e1e2. Thanks @fmateoc!

PR #126, opened 2020-07-10
Add missing DRAG.DROP event if transfer flavor is not supported

### LINKED ISSUE #125: DnD: Missing DragDrop event when dragging text into an image
## Steps to reproduce
1. Open an image in TruffleSqueak (tested with TruffleSqueak 20.2.0-dev under Win32 + Linux/X11)
2. In the host system, select some text and drag it into the image

## Expected behavior
As Squeak does not support dropping text at the moment, it would be totally okay to reject the drop such as the OSVM implementation Win32 does do it (if you try to drag some text there, the window is blocked and the system cursor displays a forbidden icon). Alternatively, this could be handled similar to the OSVM X11 implementation where a `DragDrop` event with `numFiles == 0` is generated.

## Actual behavior
In TruffleSqueak, `DragEnter` and `DragMove`s are registered as usual. But after that, not any other event is registered. As a consequence of the broken protocol, the image keeps displaying a border around the world.

## Proposed change
Generate an additional `DragDrop` event if the dragged file's flavor is not supported. See #126.

LinqLover:
Fixes #125.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/io/SqueakDisplay.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/io/SqueakDisplay.java
@@ -444,6 +444,8 @@ public final class SqueakDisplay implements SqueakDisplayInterface {
                     }
                 }
             }
+            image.dropPluginFileList = new String[0];
+            addDragEvent(DRAG.DROP, dtde.getLocation());
             dtde.rejectDrop();
         }
 


LinqLover:
@fniephaus The [coverage job](https://github.com/hpi-swa/trufflesqueak/pull/126/checks?check_run_id=859179344#step:12:51) failed with a configuration error. How can I fix that? :-)

fniephaus:
Don't worry about it, I'll fix it next week. Thank you for the contribution!

LinqLover:
@fniephaus Thanks for the fast feedback! Have a nice weekend :-)

fniephaus:
Thanks for improving TruffleSqueak. Enjoy your weekend!

PR #173, opened 2022-07-24
Odd how much they have changed in a seemingly innocuous API between 6.…

psvensson:
…0b and relase. Just weird.
This is not an ideal solution, and I would recommend to consider another way to accomplish this.


fniephaus:
Thanks! Well, `openButton:at:` used private API which is allowed to change, so I am not surprised this broke. We're planning to upgrade to the 6.0 release image right after the TruffleSqueak 22.2 release, which is due next week, so this comes at the right time. :)

fniephaus (approved):

PR #179, opened 2025-02-20
Materialize stack when more than 1000 sends deep

Sabine-Schnabeltier:
Track the entries and exits of StartContextRootNode.execute() to count the Smalltalk context depth since the resumption of the active process.

When the Smalltalk context depth exceeds 1000, cause a process switch (to the current Smalltalk context), flushing the Java stack and materializing all of the virtual contexts.


Sabine-Schnabeltier:
Refactored and resubmitted

PR #180, opened 2025-03-10
Fix minor bugs in B2D plugin

Sabine-Schnabeltier:
fetchIntegerofObject(...) was crashing trying to convert a Long directly to an int. Fixed by adding an intermediate cast to long.

loadOvallineFillleftFillrightFill(...) was incorrectly calculating the height by subtracting top from right instead of top from bottom.


Sabine-Schnabeltier:
Refactored and resubmitted.

PR #181, opened 2025-03-12
Correct operatingSystemVersion & platformProcessorType for macOS

Sabine-Schnabeltier:
Squeak is checking **platformProcessorType** for either "arm64" or "aarch64" in some places and whether "arm" is a prefix of **platformProcessorType** in others. Added a comment to that effect and made a small correction to the calculation of the macOS operating system version attribute, **operatingSystemVersion**.


fniephaus (commented):
Thanks, @Sabine-Schnabeltier! I have used this opportunity to clean up the code a bit more and align it with the attributes reported by OSVM (e.g., `aarch64` on macOS).

PR #182, opened 2025-03-12
B2D: fix oval width calculation and fetchIntegerofObject()

Sabine-Schnabeltier:
`fetchIntegerofObject(...)` was crashing trying to convert a Long directly to an int. Fixed by adding an intermediate cast to long.

`loadOvallineFillleftFillrightFill(...)` was incorrectly calculating the height by subtracting top from right instead of top from bottom.

These problems would show up when **platformProcessorType** is set to "aarch64", the Squeak preferences are set to rounded buttons and not flat widgets, and the scroll bar was at its minimum vertical size, forcing the scroll thumb to be drawn as an oval rather than a rounded rectangle.


fniephaus:
Cherry-picked via https://github.com/hpi-swa/trufflesqueak/commit/7cff18f825a1c98815ede86bb5f932569733da1e

PR #183, opened 2025-03-12
Materialize stack when more than 1000 sends deep

Sabine-Schnabeltier:
Track the entries and exits of StartContextRootNode.execute() to count the Smalltalk context depth since the resumption of the active process.

When the Smalltalk context depth exceeds 1000, cause a process switch (to the current Smalltalk context), flushing the Java stack and materializing all of the virtual Smalltalk contexts.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/StartContextRootNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/StartContextRootNode.java
@@ -31,6 +31,9 @@ public final class StartContextRootNode extends AbstractRootNode {
     @CompilationFinal private int initialPC;
     @CompilationFinal private int initialSP;
 
+    private static final int MaximumStackDepth = 1000;
+    private static int ApproximateStackDepth;
+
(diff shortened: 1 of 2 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/StartContextRootNode.java
```diff
@@ -31,6 +31,9 @@ public final class StartContextRootNode extends AbstractRootNode {
     @CompilationFinal private int initialPC;
     @CompilationFinal private int initialSP;
 
+    private static final int MaximumStackDepth = 1000;
+    private static int ApproximateStackDepth;
```
fniephaus, line 35: I'd prefer if we store these two fields in the image context. This avoids global state, which we eventually need to get rid of anyway for proper multi-context support in TruffleSqueak.

Sabine-Schnabeltier: Done; please see PR #186

Sabine-Schnabeltier:
Superseded by PR #186

PR #185, opened 2025-04-04
Restore mark bit of traversed objects before returning from someInstanceOf

Sabine-Schnabeltier:
`Behavior >> someInstance` (primitive 77) returns the first instance of the receiver's class.

In the implementation of the primitive, `someInstanceOf(...)` returns immediately upon encountering the first object of that class, leaving the enumerated objects up to that point with a different state from the rest of the objects in the object graph. This causes problems for future object graph traversals:

TS someInstance before

Changed `someInstanceOf(...)` to save the objects marked during the enumeration so that they can be unmarked before it returns.

TS someInstance after

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithClassAndHash.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithClassAndHash.java
@@ -183,6 +183,18 @@ public abstract class AbstractSqueakObjectWithClassAndHash extends AbstractSquea
         }
     }
 
+    /**
+     * @return <tt>true</tt> if already marked, <tt>false</tt> otherwise
+     */
+    public final boolean tryToUnmark(final boolean currentMarkingFlag) {
+        if (getMarkingFlag() == currentMarkingFlag) {
+            toggleMarkingFlag();
+            return true;
+        } else {
+            return false;
+        }
+    }
+
     @SuppressWarnings("unused")
     public void pointersBecomeOneWay(final Object[] from, final Object[] to) {
         // Do nothing by default.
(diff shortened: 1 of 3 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithClassAndHash.java
```diff
@@ -183,6 +183,18 @@ public final boolean tryToMark(final boolean currentMarkingFlag) {
         }
     }
 
+    /**
+     * @return <tt>true</tt> if already marked, <tt>false</tt> otherwise
+     */
+    public final boolean tryToUnmark(final boolean currentMarkingFlag) {
```
fniephaus, line 189: If I see this right, the return value isn't used. Maybe we can just inline this method.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/ObjectGraphUtils.java
```diff
@@ -106,6 +109,16 @@ private ObjectTracer(final SqueakImageContext image) {
             addObjectsFromTruffleFrames();
         }
 
+        private void cancelPendingTrace(final SqueakImageContext image, final ArrayDeque<AbstractSqueakObjectWithClassAndHash> marked) {
+            AbstractSqueakObjectWithClassAndHash currentObject;
+            // Unmark known marked objects.
+            while ((currentObject = marked.pollLast()) != null) {
```
fniephaus, line 115: I'd just use a for-loop here and let the GC clean up the marked deque. No need to remove all elements one by one.

fniephaus (commented):
Very good catch and thanks for working on a fix for this! How did you find out about this? Has it caused any observable issue running some test or application?

fniephaus:
Hope you don't mind that I addressed my own comments (see https://github.com/hpi-swa/trufflesqueak/commit/f90402bd3c497d0790bb8f7a32310d9291b29c26) and merged this into `dev`.

Sabine-Schnabeltier:
> Very good catch and thanks for working on a fix for this! How did you find out about this? Has it caused any observable issue running some test or application?

The ephemeron support needs to walk the object graph and when I was looking through examples of this, I realized that `someObjectOf` could leave an inconsistent marking state. Then verified that this would cause problems in Smalltalk `allInstances` which would give a different collection of objects depending on whether `someObject` had been used (see screenshot).

Issue #174, opened 2023-09-27
Stack overflow when evaluating factorial


Sabine-Schnabeltier:
Evaluating `15000 factorial` crashes TruffleSqueak with:
`Resource exhausted: Stack overflow`

This happens even when starting with an expanded stack:
`./trufflesqueak --vm.Xss128m`
`./trufflesqueak --vm.Xss1024m`

Sabine-Schnabeltier:
Verified this problem on Intel and Apple Silicon Macs and on native and JVM Intel Linux.

fniephaus:
It seems that running `factorial` in the interpreter overflows the stack completely, due to its recursive nature. I tried to use different values and couldn't get it to work either. I will take another look when I have some time. Interestingly though, `15000 factorial` runs fine when it is JIT-compiled. You can try this:

```smalltalk
1500 factorial.  "evaluate in workspace and then wait a few seconds for the JIT to compile #factorial"
15000 factorial. "evaluate in workspace and it works fine"
```

Sabine-Schnabeltier:
I was able to get a log and Java stack trace of this problem (see partial listing, below). It appears that the thread that is executing has a stack limit of 1024, since the stack trace is exactly 1024 lines long.

Perhaps StartContextRootNode.execute could check stack depth and materialize the sender frames to avoid the crash? Is that what happens if there is an interrupt?

```
[smalltalk::primitives] FINER: PrimFullClosureValue1NodeGen failed (arguments: a FullBlockClosure @1e0a864d (with 0 args and 1 copied values in CTX TextEditor>>#selection @440e3ce6), a Text @3e67f5f2)
[smalltalk::primitives] FINER: PrimFullClosureValue3NodeGen failed (arguments: a FullBlockClosure @fd53053 (with 1 args and 0 copied values in CTX [] in [] in TextEditor>>#selection @4527f70a), a TextColor @707b1a44, 1, 5)
[smalltalk::primitives] FINER: PrimFullClosureValue3NodeGen failed (arguments: a FullBlockClosure @7132a9dc (with 1 args and 0 copied values in CTX Text>>#removeAttributesThat: @57435801), a TextColor @707b1a44, 1, 5)
[...]

PR #186, opened 2025-04-05
Materialize stack when exceeding user-defined depth

Sabine-Schnabeltier:
Track the entries and exits of StartContextRootNode.execute() to count the Smalltalk context depth since the resumption of the active process. When the Smalltalk context depth exceeds a user-defined number, cause a process switch, flushing the Java stack and materializing all of the virtual Smalltalk contexts.

This PR adds an experimental option `--context-stack-depth=<int>` to control the maximum context stack depth. If the option is not present on the command line, the value defaults to 1000. 

Setting the option to zero will entirely disable the check and (hopefully) remove any run-time overhead.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/SqueakOptions.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/SqueakOptions.java
@@ -43,6 +43,9 @@ public final class SqueakOptions {
     @Option(name = SqueakLanguageOptions.RESOURCE_SUMMARY, category = OptionCategory.USER, stability = OptionStability.EXPERIMENTAL, help = SqueakLanguageOptions.RESOURCE_SUMMARY_HELP, usageSyntax = "false|true")//
     public static final OptionKey<Boolean> ResourceSummary = new OptionKey<>(false);
 
+    @Option(name = SqueakLanguageOptions.CONTEXT_STACK_DEPTH, category = OptionCategory.USER, stability = OptionStability.EXPERIMENTAL, help = SqueakLanguageOptions.CONTEXT_STACK_DEPTH_HELP, usageSyntax = "number")//
+    public static final OptionKey<Integer> ContextStackDepth = new OptionKey<>(1000);
+
     @Option(name = SqueakLanguageOptions.SIGNAL_INPUT_SEMAPHORE, category = OptionCategory.INTERNAL, stability = OptionStability.EXPERIMENTAL, help = SqueakLanguageOptions.SIGNAL_INPUT_SEMAPHORE_HELP, usageSyntax = "false|true")//
     public static final OptionKey<Boolean> SignalInputSemaphore = new OptionKey<>(false);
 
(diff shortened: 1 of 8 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/SqueakOptions.java
```diff
@@ -43,6 +43,9 @@ public final class SqueakOptions {
     @Option(name = SqueakLanguageOptions.RESOURCE_SUMMARY, category = OptionCategory.USER, stability = OptionStability.EXPERIMENTAL, help = SqueakLanguageOptions.RESOURCE_SUMMARY_HELP, usageSyntax = "false|true")//
     public static final OptionKey<Boolean> ResourceSummary = new OptionKey<>(false);
 
+    @Option(name = SqueakLanguageOptions.CONTEXT_STACK_DEPTH, category = OptionCategory.USER, stability = OptionStability.EXPERIMENTAL, help = SqueakLanguageOptions.CONTEXT_STACK_DEPTH_HELP, usageSyntax = "number")//
+    public static final OptionKey<Integer> ContextStackDepth = new OptionKey<>(1000);
```
fniephaus, line 47: Have you done any research/experiments to determine that 1000 is a reasonable default? We explicitly increase the max stack size in TruffleSqueak (see [here](https://github.com/hpi-swa/trufflesqueak/blob/9ce8a19c95a5f4f12528fc4b9856aa283ad4809c/mx.trufflesqueak/mx_trufflesqueak.py#L368)). I'm not sure we should remove that option, but it would be good to make sure that the default limit gets us relatively close to the stack size limit.

Sabine-Schnabeltier: When evaluating `15000 factorial` before interpreter warm-up, it crashed in the mid-2000s, so that could be an upper bound. When using Squeak (browsing, evaluating, searching, etc.), I observed stack depths mostly in the low hundreds, once it got to about 800.

Since the memory usage relationship between Smalltalk contexts and Java stack frames is unknown (factorial might use much less Java stack than other message sending patterns), I thought it would be wise to give it some headroom. I did try 2000 briefly and that seemed to work. That is the extent of the research I've done so far.

I've added the command line argument so that the limit could be adjusted or removed altogether (so that it has no overhead, if people prefer to run it that way).

By the way, I would rather prevent stack overflow in a different way. If we could have a boolean that would be set when the Java stack usage has exceeded a threshold, we would not need to count frames at all.

fniephaus: Thanks for the summary. After [removing the `-Xss` flag](https://github.com/hpi-swa/trufflesqueak/pull/186/commits/c6c361901f46b4cb4ccc45064b7b6cd72177969c), I ran a couple of experiments and it seems that `8192` still works with the default Xss value, which is `2M` on my machine. We have already observed that compiled code puts less pressure on the thread stack because it typically inlines code. So that shouldn't really be a problem either.

I'm going to merge this now. Feel free to open another ticket or PR if you are unhappy with my additional changes.

fniephaus:
We should've run the test on this. One test is consistently timing out now:
```
1) de.hpi.swa.trufflesqueak.test.SqueakSUnitTest#runSqueakTest[Passing: MemoryTests>>testFrameActivationLeak (#3123)]
java.lang.AssertionError: vvvvvvvvvvvvvvvvvv TestFailure: Test timed out vvvvvvvvvvvvvvvvvv
The time is 2025-04-09T11:40:53.107+00:00
[] in Process>>signalException:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
MemoryTests>>deepStack:
...
```

https://github.com/hpi-swa/trufflesqueak/actions/runs/14356013185/job/40245493663

fniephaus:
I took a look and the test doesn't seem to trigger materialization. In fact, I also get the timeout when I revert this PR. So the only explanation I have for this right now is it's possible that`#deepStack:` was compiled and inlined multiple levels before this test runs. Essentially the same effect described [here](https://github.com/hpi-swa/trufflesqueak/issues/174#issuecomment-1753265371).

PR #188, opened 2025-04-11
Ephemeron support, version 2

Sabine-Schnabeltier:
Cuis Smalltalk v7.3 has introduced the use of Ephemerons into the image, making it impossible to open using TruffleSqueak.

This branch introduces support for ephemeron subclasses.

Ephemeron subclasses hold strongly onto all instance variables. An ephemeron signals the VM (using the finalization semaphore) whenever the only references to the ephemeron's key are through other ephemerons that will signal. 

Tested with Cuis7.3-7036.image from [https://github.com/Cuis-Smalltalk/Cuis-Smalltalk-Dev/tree/master/CuisImage](url)


Sabine-Schnabeltier:
The last patch includes changes to interrupt handling.

Previously, the timer interrupt (managed by Delay) would prevent any lower priority interrupts and semaphores from being processed if they occurred in the same 20 ms time interval.

Now, the interrupt trigger is recomputed on exit of the CheckForInterrupts***Node execute method, guaranteeing that any postponed interrupts will still be serviced.

Also, changed CheckForInterruptsState to set the trigger when a semaphore is to be signaled, reducing the delay from signalling to trigger to zero (from 0-20 ms in the previous version).

I was considering reducing INTERRUPT_CHECKS_EVERY_N_MILLISECONDS from 20 to 2 (which appears to be the OpenSmalltalkVM default), but wanted to check with you first.

fniephaus:
> I was considering reducing INTERRUPT_CHECKS_EVERY_N_MILLISECONDS from 20 to 2 (which appears to be the OpenSmalltalkVM default), but wanted to check with you first.

Yes, we should probably do what OpenSmaltalkVM does, here. We probably shouldn't even hard-code the value but use `Smalltalk vmParameterAt: 26`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsState.java
```diff
@@ -74,39 +78,70 @@ public void shutdown() {
...
-        shouldTrigger = isActive;
+    @TruffleBoundary
+    private void createOrUpdateInterruptChecks() {
+        if (executor != null) {
+            if (interruptChecks != null && !interruptChecks.isCancelled()) {
+                interruptChecks.cancel(false);
+            }
+            interruptChecks =
+                    executor.scheduleWithFixedDelay(() -> {
+                        if (!shouldTrigger) setShouldTriggerIfNeeded();
```
fniephaus, line 89: I think we should move `!shouldTrigger` into `setShouldTriggerIfNeeded` and avoid the additional lambda. Just pass `setShouldTriggerIfNeeded` to `executor.scheduleWithFixedDelay()`.

fniephaus (commented):
Left a couple of comments. I need to take another look at some point

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsState.java
Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/accessing/SqueakObjectNewNode.java
```diff
@@ -76,6 +77,23 @@ protected static final ClassObject doClassOdd(final SqueakImageContext image, fi
...
+    }
+
+    protected static final ObjectLayout getEphemeronLayoutOrNull(final ClassObject classObject) {
+        return classObject.isEphemeronClassType() ? classObject.getLayout() : null;
+    }
+
+    @Specialization(guards = {"classObject.isEphemeronClassType()"}, replaces = "doEphemeronCached")
+    protected static final EphemeronObject doEphemeronUncached(final SqueakImageContext image, final ClassObject classObject, final int extraSize) {
+        assert extraSize == 0;
+        return new EphemeronObject(image, classObject, null);
```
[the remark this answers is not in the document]
Sabine-Schnabeltier: Is this something you would like me to put in?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
```diff
@@ -771,8 +772,9 @@ public Object execute(final VirtualFrame frame, final Object receiver) {
             }
             final SqueakImageContext image = getContext();
             final boolean hasPendingFinalizations = LogUtils.GC_IS_LOGGABLE_FINE ? hasPendingFinalizationsWithLogging(image) : hasPendingFinalizations(image);
-            if (hasPendingFinalizations) {
-                image.interrupt.setPendingFinalizations(true);
+            final boolean hasPendingEphemerons = image.containsEphemerons && ObjectGraphUtils.checkEphemerons(image);
```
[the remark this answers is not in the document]
Sabine-Schnabeltier: > If I get this right, `checkEphemerons` is now used to "implement" ephemeron semantics.

Yes!

> Is there any way to leverage Java's weak references for any of this, or do we have to do this manually similar to allInstances?

I thought about this, but Cuis' ephemeron documentation (which I copied into the Java class as a comment) says that the ephemeron holds on to its instance variables strongly and signals when the key is reachable only through paths that include ephemerons. If we held onto the key weakly, it could be garbage collected and would no longer be part of the ephemeron.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsState.java
Sabine-Schnabeltier, line 1: Addresses issue #189. By making `inActive` volatile and temporarily setting `shouldTrigger`, we can avoid a race condition where `shouldTrigger` could be set to true by the `interruptChecks` thread waking at just the right time.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsState.java
```diff
@@ -132,6 +132,8 @@ public void activate() {
     }
 
     public void deactivate() {
+        /* avoid race condition caused by interference from interruptChecks thread */
+        shouldTrigger = true;
```
fniephaus, line 136: wait, why do we set shouldTrigger to true here, and then to false?

Sabine-Schnabeltier: It may be unnecessary... I just wanted to make sure that if interruptChecks fires, it would see that shouldTrigger is already true and not potentially change shouldTrigger to true in the other thread.

src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/runCuisTests.st
```diff
@@ -19,6 +19,8 @@ failingTests := OrderedCollection new.
 {
     #FloatTest -> #(#testIsDenormalized #testPrimTruncated).
     #ProcessorTest -> #("flaky" #testGrabProcessor #testGrabProcessorOnlyForNoTimeout #testGrabProcessorOnlyForTimeout #testValueUnpreemptively).
+    #ProcessTerminateBug -> ProcessTerminateBug testSelectors.
+    #ProcessTerminateUnwindNonLocalReturn -> ProcessTerminateUnwindNonLocalReturn testSelectors.
```
fniephaus, line 23: FYI: I believe some of your changes have caused these nasty tests to fail, probably related to interrupt handling.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/accessing/SqueakObjectNewNode.java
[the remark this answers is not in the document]
fniephaus: Nothing to do here :)

Sabine-Schnabeltier:
Have you noticed that the failing tests all seem to be x86 Linux and they pass on Windows and Mac?

Could it be that the x86 Linux graalVM may be doing something different on that platform?

src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/runCuisTests.st
Sabine-Schnabeltier: When you are done for the day, I can take a look...

fniephaus: I also had to mark `ProcessTest` as failing to get to the end of the run on native. I then get [these new failures](https://github.com/hpi-swa/trufflesqueak/actions/runs/14563148672/job/40848971096#step:14:70):
```
&#16r274C; EphemeronIdentityDictionaryTest>>#testFinalizeValuesWhenLastChainContinuesAtFront has failed because: TestFailure: Assertion (7) failed. Expected 2 but was 3.
&#16r274C; FinalizationRegistryTest>>#testAddGC has failed because: TestFailure: Assertion (3) failed. 
&#16r274C; FinalizationRegistryTest>>#testEKeyDictionary has failed because: TestFailure: Assertion (1) failed. 
&#16r274C; FinalizationRegistryTest>>#testFinalizationWithError has failed because: TestFailure: Assertion (2) failed. 
&#16r274C; FinalizationTest>>#testFinalization has failed because: TestFailure: Assertion (2) failed. 
&#16r274C; FinalizationTest>>#testFinalization2 has failed because: TestFailure: Assertion (2) failed. 
```

Unfortunately, the Cuis test runner no longer prints after each test, so it's hard to find out what tests of `ProcessTest` explode. One probably needs to run these manually in the UI.

Let me continue on a different branch. Feel free to make new changes to your PR.

fniephaus: For reference: https://github.com/Cuis-Smalltalk/Cuis-Smalltalk-Dev/commit/8c1397b998e6484079c4ba090d4eeeeb95bafc65#r155630649

fniephaus (approved):
LGTM

Issue #190, opened 2025-04-16
Crash in CompiledCodeInspectorTest>>#testValuePaneModify


Sabine-Schnabeltier:
It appears that `CompiledCodeObject.hasMethodClass()` is making the assumption that the method class association is either **Association**-like or nil, but it can actually be **CompiledBlock** or **CompiledMethod**:
```
(CompiledCode allSubInstances collect: [ :cc | (cc literalAt: cc numLiterals) class]) asSet
	a Set(Association CompiledBlock CompiledMethod ClassBinding) 
```

Here is the Java stack trace for the crash:
```
Choose Smalltalk image: 2
Downloading Squeak/Smalltalk (6.0-22104)...
[trufflesqueak] Running Squeak6.0-22104-64bit.image on GraalVM CE (latency mode)...
[trufflesqueak] Image loaded in 379ms.
ERROR: java.lang.ClassCastException: class de.hpi.swa.trufflesqueak.model.CompiledCodeObject cannot be cast to class de.hpi.swa.trufflesqueak.model.AbstractPointersObject (de.hpi.swa.trufflesqueak.model.CompiledCodeObject and de.hpi.swa.trufflesqueak.model.AbstractPointersObject are in module de.hpi.swa.trufflesqueak of loader 'app')
org.graalvm.polyglot.PolyglotException: java.lang.ClassCastException: class de.hpi.swa.trufflesqueak.model.CompiledCodeObject cannot be cast to class de.hpi.swa.trufflesqueak.model.AbstractPointersObject (de.hpi.swa.trufflesqueak.model.CompiledCodeObject and de.hpi.swa.trufflesqueak.model.AbstractPointersObject are in module de.hpi.swa.trufflesqueak of loader 'app')
	at de.hpi.swa.trufflesqueak/de.hpi.swa.trufflesqueak.model.CompiledCodeObject.hasMethodClass(CompiledCodeObject.java:628)
	at de.hpi.swa.trufflesqueak/de.hpi.swa.trufflesqueak.nodes.primitives.impl.ControlPrimitivesFactory$PrimFlushCacheByMethodNodeFactory$PrimFlushCacheByMethodNodeGen.executeAndSpecialize(ControlPrimitivesFactory.java:5202)
[...]

PR #191, opened 2025-04-16
Protect access in CompiledCodeObject.hasMethodClass()

Sabine-Schnabeltier:
This pull request addresses issue #190.

`CompiledCodeObject.hasMethodClass()` was making the assumption that the method class association is either **Association**-like or nil, but it can actually be **CompiledBlock** or **CompiledMethod** if the receiver is a **CompiledBlock**.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
@@ -70,6 +70,7 @@ public final class ClassObject extends AbstractSqueakObjectWithClassAndHash {
         instanceVariables = copiedInstanceVariablesOrNull;
         organization = original.organization == null ? null : original.organization.shallowCopy();
         pointers = original.pointers.clone();
+        initializeLayout();
     }
 
     public ClassObject(final SqueakImageContext image, final ClassObject classObject, final int size) {
@@ -90,16 +91,16 @@ public final class ClassObject extends AbstractSqueakObjectWithClassAndHash {
     }
 
     public ObjectLayout getLayout() {
-        if (layout == null) {
-            CompilerDirectives.transferToInterpreterAndInvalidate();
-            layout = new ObjectLayout(this, getBasicInstanceSize());
-        }
+        assert layout != null : this + " has layout of null";
         return layout;
+    }
+
+    private void initializeLayout() {
+        layout = new ObjectLayout(getBasicInstanceSize());
     }
 
     public void updateLayout(final ObjectLayout newLayout) {
         assert layout == null || !layout.isValid() : "Old layout not invalidated";
-        CompilerDirectives.transferToInterpreterAndInvalidate();
         layout = newLayout;
     }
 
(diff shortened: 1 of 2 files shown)


fniephaus (approved):
LGTM, thanks for the fix!

fniephaus:
Can we mark the [test as passing](https://github.com/hpi-swa/trufflesqueak/blob/62fac6143be165b034ad0d66228d86533c0e13d4/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/tests.properties#L837) now?

Sabine-Schnabeltier:
Yes!

Sabine-Schnabeltier:
Second patch addresses #192, `ProtoObject >> tryPrimitive:WithArgs: crashes when send with wrong number of arguments`

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
```diff
@@ -669,7 +669,12 @@ protected static final AbstractPrimitiveNode createPrimitiveNode(final long prim
 
         protected static final Object primitiveWithArgs(final VirtualFrame frame, final Object receiver, final ArrayObject argumentArray,
                         final AbstractPrimitiveNode primitiveNode, final ArrayObjectToObjectArrayCopyNode toObjectArrayNode, final Node inlineTarget) {
-            return primitiveNode.executeWithArguments(frame, receiver, toObjectArrayNode.execute(inlineTarget, argumentArray));
+            try {
+                return primitiveNode.executeWithArguments(frame, receiver, toObjectArrayNode.execute(inlineTarget, argumentArray));
+            } catch (NegativeArraySizeException | ArrayIndexOutOfBoundsException oob) {
```
fniephaus, line 674: I've looked into this and it seems you have fixed the symptoms but not the root cause. The problem appears to be that the nodes annotated with `@SqueakPrimitive(indices = 83)` don't check the argument count during dispatch.

`Interpreter>>#primitivePerform` does this:

```smalltalk
"Only test CompiledMethods for argument count - other objects will have to take their chances"
(self isOopCompiledMethod: newMethod)
	ifTrue: [self success: (self argumentCountOf: newMethod) = argumentCount].
```

I'd like us to fix the root cause instead but it's not trivial because the perform nodes reuse the `DispatchXNode` nodes and they don't check the argument count. One could derive the argument count from the selector, but that is not semantically identical to what OpenSmalltalkVM does.

fniephaus:
Apologies, I made a mistake editing your branch and accidentally closed the PR. Because of this, I no longer have permissions to edit your branch. I only wanted to drop https://github.com/hpi-swa/trufflesqueak/commit/909c78f4477f0bcc97e8eb5f1c97b0716ae186bd for now and push https://github.com/hpi-swa/trufflesqueak/commit/abdab9db265f6d751600765c4011131e5b675c18. I have done this on a separate branch and if all tests pass, I will merge the fix into main.

Let's follow up on https://github.com/hpi-swa/trufflesqueak/pull/191#discussion_r2051103169 in a separate PR.

PR #194, opened 2025-05-14
Signal all semaphores on check-for-interrupts & switch to highest priority process

Sabine-Schnabeltier:
The current check-for-interrupts code exits immediately when a semaphore for a higher priority process is signalled, leaving the other pending signals to be handled on a subsequent poll.

This delays the signalling of finalization and external semaphores when Delays are expiring frequently.

This patch signals all of the pending semaphores and catches any thrown ProcessSwitches. Since the process switch has already updated Processor's priority, the only time there would be more than one ProcessSwitch thrown would be when the new process has higher priority than the process associated with the earlier ProcessSwitch. After signalling all needed semaphores, we throw the saved ProcessSwitch (if any).

This fixes the following Cuis Ephemeron and finalization tests:

```
EphemeronIdentityDictionaryTest>>#testFinalizeValuesWhenLastChainContinuesAtFront has failed because: TestFailure: Assertion (7) failed. Expected 2 but was 3.
FinalizationRegistryTest>>#testAddGC has failed because: TestFailure: Assertion (3) failed. 

FinalizationRegistryTest>>#testEKeyDictionary has failed because: TestFailure: Assertion (1) failed. 

FinalizationTest>>#testFinalization has failed because: TestFailure: Assertion (2) failed. 
FinalizationTest>>#testFinalization2 has failed because: TestFailure: Assertion (2) failed. 
```


Sabine-Schnabeltier:
Sorry for the 5 extra commits at the top -- I couldn't figure out how to get rid of them.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsFullNode.java
```diff
@@ -36,24 +37,48 @@ public void execute(final VirtualFrame frame) {
...
             final ArrayObject externalObjects = (ArrayObject) specialObjects[SPECIAL_OBJECT.EXTERNAL_OBJECTS_ARRAY];
             if (!externalObjects.isEmptyType()) { // signal external semaphores
                 final Object[] semaphores = externalObjects.getObjectStorage();
                 Integer semaIndex;
                 while ((semaIndex = istate.nextSemaphoreToSignal()) != null) {
-                    signalSemaporeNode.executeSignal(frame, this, semaphores[semaIndex - 1]);
+                    try {
+                        signalSemaporeNode.executeSignal(frame, this, semaphores[semaIndex - 1]);
+                    } catch (final ProcessSwitch ps) {
+                        pendingSwitch = ps;
```
fniephaus, line 71: Catching `ProcessSwitch` is a neat trick to achieve the right behavior. However, in the worst case, a lot of exceptions are thrown and given the check is supposed to be as cheap as possible, I think it'd make sense if we could change the implementation, so that the new `activeContext` is returned rather than thrown as part of a `ProcessSwitch`.

Is this something you'd like to look into, or should I take a look?

fniephaus:
Could you please remove the tests that now pass from the [list of failing tests](https://github.com/hpi-swa/trufflesqueak/blob/76857168cf8afbdae6018a31913753f975e42924/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/runCuisTests.st#L21-L29)?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsFullNode.java
Sabine-Schnabeltier: I will take a look...

src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/runCuisTests.st
```diff
@@ -9,7 +9,6 @@ nonTerminatingTestCases := OrderedCollection new.
 {
     #JpegTest -> #(#test16bpp61 #test16bpp63 #test16bppLE #test16bppLE61 #test16bppLE62 #test16bppLE63 #test32bpp #test32bpp63 #test32bppLE63 #testBA32bpp #testBA32bpp63 "all previous tests pass on TruffleSqueak in JVM mode, the rest is not yet supported" #test16bpp #test16bpp62 #testBA16bpp #testBA16bpp61 #testBA16bpp62 #testBA16bpp63 #testBA16bppLE #testBA16bppLE61 #testBA16bppLE62 #testBA16bppLE63 #testGray #testGray61 #testGray62 #testGray63 #testGrayBA #testGrayBA61 #testGrayBA62 #testGrayBA63).
     #ProcessTest -> #(#testResumeWithEnsureAfterBCR #testTerminateInEnsure).
-    #SemaphoreTest -> #("unexpected null reads" #testSemaInCriticalWaitInEnsure #testSemaInCriticalWaitInEnsure2).
```
fniephaus, line 12: Looks like this still causes an assertion error that we haven't fixed yet: https://github.com/hpi-swa/trufflesqueak/actions/runs/15034462552/job/42263054508#step:14:817

Maybe revert the line?

Sabine-Schnabeltier:
I'll fix up the style failures when I send the new version of the patch -- hopefully in a few hours.

fniephaus (approved):
LGTM, thank you!

PR #200, opened 2025-06-26
Make ContextPart>>terminateTo: confirm sender chain before terminating Contexts

Sabine-Schnabeltier:
1. PrimTerminateNode describes primitive 196 as
```
            /*
             * Terminate all the Contexts between me and previousContext, if previousContext is on
             * my Context stack. Make previousContext my sender.
             */
```
but the implementation was terminating Contexts above the receiver, stopping when finding either previousContext or nil.

Modified implementation to check for previousContext first, and, if found, then do the termination.

2. Corrected an asymmetry between ContextObject.terminate() and FrameAccess.terminate(): ContextObject cleared the modified sender flag, but FrameAccess did not. Not sure ContextObject.clearModifiedSender needs to check before clearing the flag [just following the example in ContextObject.removeSender()]

3. Fixed typo and added some documentation for the return bytecodes.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
@@ -209,14 +209,14 @@ public final class ContextObject extends AbstractSqueakObjectWithClassAndHash {
     public AbstractSqueakObject getSender() {
         final Object value = getFrameSender();
         if (value instanceof final FrameMarker f) {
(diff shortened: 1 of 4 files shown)


Sabine-Schnabeltier:
`How did you notice the incorrect behavior? Is there something we can do to ensure we don't regress (like adding a test case)?`

I've been trying to track down a problem in non-local returns and noticed most of these issues while reading the source. (Unfortunately, none of these changes fixes the problem I've been tracking.)

The problem with HandleNonLocalReturnNode was discovered when manually running all of the TruffleSqueak tests found in `3) TruffleSqueak test image (6.0-22104)`.

PR #201, opened 2025-07-26
Separate the roles of NLR and NVR

Sabine-Schnabeltier:
Three exceptions are used to alter control flow; the first is for non-local returns (up arrows in blocks, or block returns), the second is for normal returns (method return or block return to caller) when the sender has been modified since the Context was started; the third is used to exit the interpreter loop with a return value.

- `NonLocalReturn`:  passes a return value to a sender, executing unwind blocks found between current Context and sender Context
- `NonVirtualReturn`: passes a return value to a Context that is not immediately above the current Java execution node
- `TopLevelReturn`:  passes a return value to the `ExecuteTopLevelContextNode` at the top of the Java stack

This PR includes the following changes:

1. `ResumeContextRootNode` now clears the modified sender flag of the Context being resumed and `commonNLReturn()` in `ExecuteTopLevelContextNode` clears the modified sender flag before `executeAboutToReturn`. They do this so that the modified sender field of a Context correctly indicates whether the Truffle execution stack is different from the Context sender chain.
2. Changed `ReturnFromMethodNode` and `AbstractBlockReturnNode` to use `NonVirtualReturn` when they have modified senders.
3. Changed `AbstractBlockReturnNode` to return to the caller rather than to the sender of the home Context.
4. In `ExecuteTopLevelContextNode`, `NonLocalReturn` and `NonVirtualReturn` were both handled in similar ways: they both would process unwind-marked methods and terminate Contexts found on the sender chain. Changed this so that only `NonLocalReturn` terminates Contexts and handles unwind-marked methods.
5. In `ExecuteTopLevelContextNode`, try to separate errors (returning to a `nil` sender) from exiting the interpreter (`ERROR: A Squeak/Smalltalk image cannot return a result, it can only exit.`)

Known issues:

1. At this time, the internal virtualization of `aboutToReturn:` does not handle the possibility of a `ProcessSwitch` or error occurring during the execution of the `ensure:` block. Some mechanism needs to be created to make it possible to resume the interrupted non-local return.
2. There is a possibility that a non-local return may encounter frames with modified senders and would subsequently need to skip Truffle stack frames before resuming the unwind operation.
3. Sending Smalltalk messages from `ExecuteTopLevelContextNode` can confuse the Truffle-stack-frame-walking loops that continue following the Context sender chain after encountering a `ResumeContextRootNode`.
4. Some of the Cuis ProcessTests are assuming that non-local returns are checking for the return Context being on the sender chain prior to doing any unwinding. Among these tests is `ProcessTest>>testResumeWithEnsureAfterBCR`

The following tests pass only after several retries:
```
StringTest>>#testFindSelector
SHParserST80Test>>#testNumbers
```

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
@@ -76,8 +76,8 @@ public abstract class AboutToReturnNode extends AbstractNode {
         @Specialization(guards = {"hasModifiedSender(frame)"})
         protected static final void doAboutToReturn(final VirtualFrame frame, final NonLocalReturn nlr,
                         @Cached("createAboutToReturnSend()") final Dispatch2Node sendAboutToReturnNode) {
-            assert nlr.getTargetContextOrMarker() instanceof ContextObject;
-            sendAboutToReturnNode.execute(frame, FrameAccess.getContext(frame), nlr.getReturnValue(), nlr.getTargetContextOrMarker());
+//            assert nlr.getTargetContextOrMarker() instanceof ContextObject;
+            sendAboutToReturnNode.execute(frame, FrameAccess.getContext(frame), nlr.getReturnValue(), nlr.getOrCreateTargetContext());
         }
     }
 
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
@@ -77,6 +77,7 @@ public abstract class AboutToReturnNode extends AbstractNode {
         @Specialization(guards = {"hasModifiedSender(frame)"})
         protected static final void doAboutToReturn(final VirtualFrame frame, final NonLocalReturn nlr,
                         @Cached("createAboutToReturnSend()") final Dispatch2Node sendAboutToReturnNode) {
+            // @formatter:off
             /*
              *  aboutToReturn: result through: firstUnwindContext
 	         *      "Called from VM when an unwindBlock is found between self and its home.
@@ -84,6 +85,7 @@ public abstract class AboutToReturnNode extends AbstractNode {
              *
 	         *      self methodReturnContext return: result through: firstUnwindContext
              */
+            // @formatter:on
             // Message receiver should be home Context to return from.
             // Last argument should be the first unwind-marked Context or nil.
             sendAboutToReturnNode.execute(frame, nlr.getHomeContext(), nlr.getReturnValue(), NilObject.SINGLETON);
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
@@ -80,10 +80,10 @@ public abstract class AboutToReturnNode extends AbstractNode {
             // @formatter:off
             /*
              *  aboutToReturn: result through: firstUnwindContext
-	         *      "Called from VM when an unwindBlock is found between self and its home.
-	         *      Return to home's sender, executing unwind blocks on the way."
+             *      "Called from VM when an unwindBlock is found between self and its home.
+             *      Return to home's sender, executing unwind blocks on the way."
              *
-	         *      self methodReturnContext return: result through: firstUnwindContext
+             *      self methodReturnContext return: result through: firstUnwindContext
              */
             // @formatter:on
             // Message receiver should be home Context to return from.
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
@@ -76,7 +76,7 @@ public abstract class AboutToReturnNode extends AbstractNode {
         @Specialization(guards = {"hasModifiedSender(frame)"})
         protected static final void doAboutToReturn(final VirtualFrame frame, final NonLocalReturn nlr,
                         @Cached("createAboutToReturnSend()") final Dispatch2Node sendAboutToReturnNode) {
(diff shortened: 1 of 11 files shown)


Sabine-Schnabeltier:
An additional transient error is:

`StringTest>>#testFindSelector has errored because: MessageNotUnderstood: MethodContext>>isEmpty`

(All referenced tests are from `runCuisTests.st`

fniephaus:
The diff of this PR has become a bit overwhelming. Could you maybe rebase your changes onto main to reduce the diff to what you have changed?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithClassAndHash.java
```diff
@@ -145,7 +154,8 @@ public final boolean needsSqueakHash() {
     }
 
     public final void setSqueakHash(final int newHash) {
-        assert newHash <= SqueakImageConstants.IDENTITY_HASH_HALF_WORD_MASK;
+        assert assertNotForwarded();
+        assert HASH_UNINITIALIZED <= newHash && newHash <= SqueakImageConstants.IDENTITY_HASH_HALF_WORD_MASK;
```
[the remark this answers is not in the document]
Sabine-Schnabeltier: No, just saw that the newHash is signed and the check was only for it to be too large.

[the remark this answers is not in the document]
fniephaus: Ok, good point. I hope you don't mind I changed the assertion in a different way: https://github.com/hpi-swa/trufflesqueak/commit/3e94e17f89824cd9b0e26bc6e67ac43f2b252630.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ExecuteTopLevelContextNode.java
```diff
@@ -133,82 +140,76 @@ private static ContextObject returnTo(final ContextObject activeContext, final A
...
-            assert !context.isPrimitiveContext();
             if (context.getCodeObject().isUnwindMarked()) {
-                assert !context.hasClosure();
-                /* "context is marked; break out" */
-                return sendAboutToReturn(startContext, returnValue, context);
+                try {
+                    // TODO: make this better. Clearing the modified sender permits virtualization
+                    // of aboutToReturn
+                    context.clearModifiedSender();
+                    AboutToReturnNode.create(context.getCodeObject()).executeAboutToReturn(context.getTruffleFrame(), nlr);
```
fniephaus, line 202: `commonNLReturn()` is behind a `@TruffleBoundary` so excluded from compilation where values may be virtualized. That's also why we previously used `sendAboutToReturnNode` here. Were you trying to optimize this here or can we use `sendAboutToReturn()` and the node again?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ExecuteBytecodeNode.java
```diff
@@ -54,6 +55,10 @@ public Object execute(final VirtualFrame frame, final int startPC) {
             return interpretBytecode(frame, startPC);
         } catch (final NonLocalReturn nlr) {
             /* {@link getHandleNonLocalReturnNode()} acts as {@link BranchProfile} */
+            if (FrameAccess.isDead(frame)) {
+                LogUtils.SCHEDULING.fine("ExecuteBytecodeNode: encountered dead frame during NLR");
+                throw nlr;
+            }
             return getHandleNonLocalReturnNode().executeHandle(frame, nlr);
```
fniephaus, line 62: The comment means that the lazy-initialized node is used as a `BranchProfile` to hide the code path if it has never been taken for a specific method. The new code here, however, is before the lazy-init getter. To hide it also, you have to do something like this instead:

```suggestion
            /* {@link getHandleNonLocalReturnNode()} acts as {@link BranchProfile} */
            final HandleNonLocalReturnNode handleNode = getHandleNonLocalReturnNode();
            if (FrameAccess.isDead(frame)) {
                LogUtils.SCHEDULING.fine("ExecuteBytecodeNode: encountered dead frame during NLR");
                throw nlr;
            }
            return handleNode.executeHandle(frame, nlr);
```

Hope this makes sense :)

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
```diff
@@ -76,8 +76,8 @@ protected final void doAboutToReturnVirtualizedNothing(final VirtualFrame frame,
         @Specialization(guards = {"hasModifiedSender(frame)"})
         protected static final void doAboutToReturn(final VirtualFrame frame, final NonLocalReturn nlr,
                         @Cached("createAboutToReturnSend()") final Dispatch2Node sendAboutToReturnNode) {
-            assert nlr.getTargetContextOrMarker() instanceof ContextObject;
-            sendAboutToReturnNode.execute(frame, FrameAccess.getContext(frame), nlr.getReturnValue(), nlr.getTargetContextOrMarker());
+            /* assert nlr.getTargetContextOrMarker() instanceof ContextObject; */
```
fniephaus, line 79: If this no longer holds true, should we remove it?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ExecuteTopLevelContextNode.java
Sabine-Schnabeltier: What I was trying to do here was to force the use of AboutToReturnNode>>doAboutToReturnVirtualized because I was unsure whether we could mix the virtualized and message-based implementations of aboutToReturn.

It seems to me that once we start using doAboutToReturnVirtualized, the homeContext no longer has a sender and so the message would fail.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/AboutToReturnNode.java
Sabine-Schnabeltier: Yes!

Sabine-Schnabeltier:
Sorry for the duplicated commits :( One of these days I will learn the proper way to rebase and push in IntelliJ....

Sabine-Schnabeltier:
Replaced by #209

PR #208, opened 2025-08-24
save PC before executeReturn

Sabine-Schnabeltier:
Save the current PC in the frame before executing a return byte code since there is the possibility that aboutToReturn or cannotReturn would be sent and, in response to the message, the image might try to examine or resume the current context.


fniephaus (approved):
Merged via https://github.com/hpi-swa/trufflesqueak/commit/f4301ce3758998d6b1ab141cf52655ddec30325d

PR #210, opened 2025-09-06
Implement Image flag bit for "Process preemption yields"

Sabine-Schnabeltier:
The image flags for Squeak and Cuis include a bit controlling whether, when a Process gets preempted, it should yield (added to the end of the runnable Processes list for its priority) or not (added to the front of the list).

The original Blue Book implementation was for preempted Processes to yield; both Squeak and Cuis have the image flag bit set so that preempted Processes do not yield.


fniephaus (approved):
LGTM, thank for the important fix!

PR #213, opened 2025-09-22
Return-top bytecodes do not pop

Sabine-Schnabeltier:
The OpenSmalltalkVM return-top bytecode implementations fetch the top of the stack, rather than pop the top of the stack.

However, this causes one of the WeakRegistry tests to fail in TS: the Object that the registry is using as a key ought to be garbage collected, but instead it is being held onto by a reference in a dead Context. Not sure how the dead Context avoids being garbage collected though...

An additional twist to this is that, within the image, the Context simulation of return-top pops the stack, meaning the OpenSmalltalkVM implementation is wrong.

Technically, the TS implementation is correct. But I have a suspicion that one of the CuisTests (ProcessTest>>#testResumeWithEnsureAfterBCR) may be relying on the value still being on the stack, still checking.

Not clear if this is worth pursuing, but I thought you should be aware of this difference.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
@@ -438,7 +438,6 @@ public final class ControlPrimitives extends AbstractPrimitiveFactoryHolder {
                 if (myContextOrNil instanceof final ContextObject context) {
                     assert !context.isDead() : "Have Context but it is dead: " + context;
                 } else {
-                    // If it is NOT a ContextObject, assert that it must be the NilObject.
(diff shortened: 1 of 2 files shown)


fniephaus:
Thanks for the PR, @Sabine-Schnabeltier. It seems `WeakRegistryTest>>#testFinalizationWithMultipleFinalizersPerObject` is now failing due to this change. Could you take a look and try to find out why?

Sabine-Schnabeltier:
The Object added as the key to the WeakRegistry ought to be garbage collected, but instead it is being held onto by a stack reference in a dead Context. Before this PR, the pop would have (eventually) nilled that stack slot and so the reference would go away even though the Context was still around.

 Not sure how the dead Context is still alive -- any suggestions for how to find the reference?

Sabine-Schnabeltier:
From Squeak Context class comment:
`When a context is returned from, its sender and pc are nilled, and, if the context is still referred to, the virtual machine guarantees to preserve only the arguments after a return.`

Sabine-Schnabeltier:
Thanks!

PR #249, opened 2026-04-04
prevent access to out of bounds stack slots

Sabine-Schnabeltier:
The bounds check on Context stack access is too permissive and can crash the interpreter. From the comments in the Smalltalk source code, it appears that the access primitives should fail if accessing slots above the stack pointer. It should also fail if accessing slots less than 1.

Since we now make smaller stack frames than requested by the CompiledMethod flags, we may need to revisit the primitive that sets the stack pointer to resize the stack frame if the stack pointer is moved algorithmically to a value larger than the deepest stack the method's evaluation would reach.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
@@ -149,7 +149,7 @@ public class ContextPrimitives extends AbstractPrimitiveFactoryHolder {
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 210)
     protected abstract static class PrimContextAtNode extends AbstractPrimitiveNode implements Primitive1WithFallback {
-        @Specialization(guards = {"index <= receiver.getStackPointer()"})
+        @Specialization(guards = {"1 <= index", "index <= receiver.getStackPointer()"})
         protected static final Object doContextObject(final ContextObject receiver, final long index,
                         @Bind final Node node,
                         @Cached final ContextObjectReadNode readNode) {
@@ -160,7 +160,7 @@ public class ContextPrimitives extends AbstractPrimitiveFactoryHolder {
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 211)
     protected abstract static class PrimContextAtPutNode extends AbstractPrimitiveNode implements Primitive2WithFallback {
-        @Specialization(guards = "index <= receiver.getStackPointer()")
+        @Specialization(guards = {"1 <= index", "index <= receiver.getStackPointer()"})
         protected static final Object doContextObject(final ContextObject receiver, final long index, final Object value,
                         @Bind final Node node,
                         @Cached final ContextObjectWriteNode writeNode) {
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
@@ -149,7 +149,7 @@ public class ContextPrimitives extends AbstractPrimitiveFactoryHolder {
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 210)
     protected abstract static class PrimContextAtNode extends AbstractPrimitiveNode implements Primitive1WithFallback {
-        @Specialization(guards = {"index <= receiver.size()"})
+        @Specialization(guards = {"index <= receiver.getStackPointer()"})
         protected static final Object doContextObject(final ContextObject receiver, final long index,
                         @Bind final Node node,
                         @Cached final ContextObjectReadNode readNode) {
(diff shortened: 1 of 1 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -160,7 +160,7 @@ protected static final Object doContextObject(final ContextObject receiver, fina
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 211)
     protected abstract static class PrimContextAtPutNode extends AbstractPrimitiveNode implements Primitive2WithFallback {
-        @Specialization(guards = "index <= receiver.size()")
+        @Specialization(guards = {"1 <= index", "index <= receiver.getStackPointer()"})
```
fniephaus, line 163: AFAIC, it is allowed to write into slots past the stack pointer. Right now, it would through an IIOB or so, but for writes above the calculated max stack size, we should probably write into an aux slot. So `index <= receiver.getStackPointer()` should probably be `index <= receiver.getCodeObject().getMaxNumStackSlots()`, and then we need another specializations for `getCodeObject().getMaxNumStackSlots() <= receiver.size()`.

Or do you disagree? :)

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -172,13 +172,20 @@ protected static final Object doContextObject(final ContextObject receiver, fina
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 212)
     protected abstract static class PrimContextSizeNode extends AbstractPrimitiveNode implements Primitive0WithFallback {
+        /*
+         * "Note that it is impossible to determine the real object size of a Context except by
```
fniephaus, line 176: Mh, now that I think about this, should this primitive always return `getCodeObject().getSqueakContextSize()`?

Sabine-Schnabeltier: From `StackInterpreterPrimitives`:
```
primitiveContextSize
	"Special version of primitiveSize for accessing contexts.
	 Written to be varargs for use from mirror primitives."
	| rcvr sz hdr fmt totalLength fixedFields |
	<inline: false>
	rcvr := self stackTop.
	hdr := objectMemory baseHeader: rcvr.
	fmt := objectMemory formatOfHeader: hdr.
	totalLength := objectMemory lengthOf: rcvr baseHeader: hdr format: fmt.
	fixedFields := objectMemory fixedFieldsOf: rcvr format: fmt length: totalLength.
	(objectMemory isContextHeader: hdr)
		ifTrue:
			[self externalWriteBackHeadFramePointers.
			sz := self stackPointerForMaybeMarriedContext: rcvr]
		ifFalse: [sz := totalLength - fixedFields].
	self methodReturnInteger: sz
```
If the receiver is a `BlockContext` or a `MethodContext`, it returns the stack pointer value; otherwise, it returns the standard indexable variables count.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: From StackInterpreterPrimitives:
```
primitiveContextAt
	"Special version of primitiveAt for accessing contexts.
	 Written to be varargs for use from mirror primitives."
	| index value aContext spouseFP hdr fmt totalLength fixedFields stSize |
	<inline: false>
	<var: #spouseFP type: #'char *'>
	index := self stackTop.
	(objectMemory isIntegerObject: index) ifFalse:
		[self primitiveFailFor: PrimErrBadArgument.
		 ^self].
	index := objectMemory integerValueOf: index.
	aContext := self stackValue: 1.
	"Duplicating much of stObject:at:put: here allows stObject:at:put: to omit tests for contexts."
	hdr := objectMemory baseHeader: aContext.
	(objectMemory isContextHeader: hdr) ifFalse: "might be an instance of a subclass"
		[value := self stObject: aContext at: index.
		 self successful ifTrue:
			[self pop: argumentCount + 1 thenPush: value].
		 ^self].
	self externalWriteBackHeadFramePointers.
	(self isStillMarriedContext: aContext) ifFalse:
		[fmt := objectMemory formatOfHeader: hdr.
		 totalLength := objectMemory lengthOf: aContext baseHeader: hdr format: fmt.
		 fixedFields := objectMemory fixedFieldsOf: aContext format: fmt length: totalLength.
		 stSize := self fetchStackPointerOf: aContext.
		 (index between: 1 and: stSize) ifFalse:
			[self primitiveFailFor: PrimErrBadIndex.
			 ^self].			
		value := self subscript: aContext with: (index + fixedFields) format: fmt.
		self pop: argumentCount + 1 thenPush: value.
		^self].
	spouseFP := self frameOfMarriedContext: aContext.
	(index between: 1 and: (self stackPointerIndexForFrame: spouseFP)) ifFalse:
		[self primitiveFailFor: PrimErrBadIndex.
		 ^self].
	value := self temporary: index - 1 in: spouseFP.
	self pop: argumentCount + 1 thenPush: value
```

If the receiver is a `Context` or a `BlockContext`, the bounds check is between 1 and stack pointer; otherwise, it uses `#stObject:at:`. Similarly for `#primitiveContextAtPut` in the same class.

Sabine-Schnabeltier:
When resizing the frame via a change to the stack pointer (which should be exceptionally rare), we could just make the frame the size specified by the `CompiledMethod` flags. This would allow us to avoid checking the auxSlots in other parts of the interpreter.

fniephaus:
> we could just make the frame the size specified by the CompiledMethod flags

The slots arrays are final so we cannot resize an existing frame. That's why we use aux slots. We can only resize the number of slots on the descriptor so that future frame have a larger size.

Sabine-Schnabeltier:
> The slots arrays are final so we cannot resize an existing frame. That's why we use aux slots. We can only resize the number of slots on the descriptor so that future frame have a larger size.

I can't seem to find where any aux slots are set. Does it happen automatically, behind the scenes, in Truffle?

Sabine-Schnabeltier:
Superseded by #250

PR #250, opened 2026-04-10
Eliminate nilling when popping stack; nil unreachable stack before GC

Sabine-Schnabeltier:
### Summary
Eliminate the eager nilling of stack slots in `AbstractInterpreterNode.pop()` and `popN()`, and instead hide references to objects in stack slots above the stack pointer during object graph traversals in `ObjectGraphUtils`. The `FrameAccess` stack iteration methods were modified to explicitly stop iteration at the stack pointer limit. This transition to lazy nilling aligns the behavior with the Open Smalltalk VM and yields an approximate 4% speedup on benchmarks when executing on the JVM.

### GC & Finalization Determinism
Even though objects referenced from "dead" stack slots are invisible to Smalltalk, they will persist in the JVM. 
* The garbage collection primitives were modified to nil these "dead" stack slots before performing the GC (via `flushDeadReferences()`). 
* To make finalization deterministic, the GC primitives were modified to check for interrupts and handle them.

### Object Graph Traversal Policies
* Object graph traversals in `ObjectGraphUtils` were modified to indicate whether they want to `SCAN` or `SCRUB` frames they encounter. 
* The addition of `flushDeadReferences()` meant that `checkEphemerons()` could be simplified to safely ignore `becomeMap`. 
* Because of a possible (but very unlikely) race condition, some of the processing of ephemerons was moved from the parallel tracing tasks to the sequential trace in the second half of `checkEphemerons()`.

### Primitive and Node Updates
* **ControlPrimitives:** All methods that could cause a process switch were modified to push the result onto the stack rather than simply bumping the stack pointer. With the new lazy nilling semantics, merely bumping the stack pointer would leave a residual value (likely the message receiver) as the apparent result of the primitive, whereas the previous eager nil-on-pop behavior left `nil`.
* **ContextPrimitives:** The changes from #249 were incorporated, with an additional change to nil newly hidden or exposed stack slots when the stack pointer is manually modified.
* **Interrupts:** `CheckForInterruptsInLoopNode` was modified to save the current stack pointer in the frame in the event an interrupt occurs during a backward branch.
* **ResumeContextRootNode:** `execute()` was modified to save `activeContext` in its frame argument so that `iterateFrames()` can successfully find the context when execution is taking place directly within that active context.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
@@ -37,8 +37,10 @@ public final class ResumeContextRootNode extends AbstractRootNode {
     @Override
     public Object execute(final VirtualFrame frame) {
         assert !activeContext.isDead() : "Terminated contexts cannot be resumed";
-        // Make it so ObjectGraphUtils can find the active context when executing
-        // strictly within the active context.
+        /*
+         * Make it so ObjectGraphUtils can find the active context when executing strictly within
+         * the active context.
+         */
         FrameAccess.setContext(frame, activeContext);
         activeContext.clearModifiedSender();
         final int pc = instructionPointerProfile.profile(activeContext.getInstructionPointerForBytecodeLoop());
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
@@ -15,6 +15,7 @@ import com.oracle.truffle.api.profiles.IntValueProfile;
 
 import de.hpi.swa.trufflesqueak.image.SqueakImageContext;
 import de.hpi.swa.trufflesqueak.model.ContextObject;
+import de.hpi.swa.trufflesqueak.util.FrameAccess;
 
 public final class ResumeContextRootNode extends AbstractRootNode {
     private ContextObject activeContext;
@@ -36,6 +37,9 @@ public final class ResumeContextRootNode extends AbstractRootNode {
     @Override
     public Object execute(final VirtualFrame frame) {
         assert !activeContext.isDead() : "Terminated contexts cannot be resumed";
+        // Make it so ObjectGraphUtils can find the active context when executing
+        // strictly within the active context.
+        FrameAccess.setContext(frame, activeContext);
         activeContext.clearModifiedSender();
         final int pc = instructionPointerProfile.profile(activeContext.getInstructionPointerForBytecodeLoop());
         final int sp = stackPointerProfile.profile(activeContext.getStackPointer());
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -370,7 +370,6 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
     }
 
     protected static final Object pop(final VirtualFrame frame, final int sp) {
-        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
         final int slotIndex = FrameAccess.getStackStart() + sp;
         return frame.getObjectStatic(slotIndex);
     }
@@ -385,7 +384,6 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
 
     @ExplodeLoop
     private static Object[] popNExploded(final VirtualFrame frame, final int sp, final int numPop) {
-        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
         final int firstSlotIndex = FrameAccess.getStackStart() + sp - numPop;
         final Object[] stackValues = new Object[numPop];
         for (int i = 0; i < numPop; i++) {
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -295,7 +295,7 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         if (homeContext.canBeReturnedTo()) {
             final ContextObject firstMarkedContext = firstUnwindMarkedOrThrowNLR(FrameAccess.getSender(frame), homeContext, result);
             if (firstMarkedContext != null) {
-                externalizePCAndSP(frame, pc, sp);
+                FrameAccess.externalizePCAndSP(frame, pc, sp);
                 if (data[currentPC] == null) {
                     CompilerDirectives.transferToInterpreterAndInvalidate();
                     data[currentPC] = insert(Dispatch2NodeGen.create(getContext().aboutToReturnSelector));
@@ -303,7 +303,7 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
                 ((Dispatch2NodeGen) data[currentPC]).execute(frame, getOrCreateContext(frame, currentPC), result, firstMarkedContext);
             }
         }
-        throw cannotReturn(frame, result);
+        throw cannotReturn(frame, pc, sp, result);
     }
 
     /**
@@ -336,9 +336,10 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         return null;
     }
 
-    private static CannotReturnToTarget cannotReturn(final VirtualFrame frame, final Object returnValue) {
+    private static CannotReturnToTarget cannotReturn(final VirtualFrame frame, final int pc, final int sp, final Object returnValue) {
         CompilerDirectives.transferToInterpreter();
         LogUtils.SCHEDULING.info("sendCannotReturn");
+        FrameAccess.externalizePCAndSP(frame, pc, sp);
         throw new CannotReturnToTarget(returnValue, GetOrCreateContextWithFrameNode.executeUncached(frame));
     }
 
@@ -397,21 +398,6 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         return FrameAccess.getStackValue(frame, sp - 1);
     }
 
-    protected static final void externalizePCAndSP(final VirtualFrame frame, final int pc, final int sp) {
-        FrameAccess.setInstructionPointer(frame, pc);
-        FrameAccess.setStackPointer(frame, sp);
-    }
-
-    protected static final int internalizePC(final VirtualFrame frame, final int pc) {
-        final int framePC = FrameAccess.getInstructionPointer(frame);
-        if (pc != framePC) {
-            CompilerDirectives.transferToInterpreter();
-            return framePC;
-        } else {
-            return pc;
-        }
-    }
-
     protected final Object getErrorObject() {
         final SqueakImageContext image = getContext();
         final int primFailCode = image.getPrimFailCode();
@@ -425,7 +411,7 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
 
     protected final void sendMustBeBooleanInInterpreter(final VirtualFrame frame, final int pc, final int sp, final Object stackValue) {
         CompilerDirectives.transferToInterpreter();
-        externalizePCAndSP(frame, pc, sp);
+        FrameAccess.externalizePCAndSP(frame, pc, sp);
         final SqueakImageContext image = getContext();
         image.mustBeBooleanSelector.executeAsSymbolSlow(image, frame, stackValue);
         throw SqueakException.create("Should not be reached");
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -37,7 +37,6 @@ import de.hpi.swa.trufflesqueak.model.ArrayObject;
 import de.hpi.swa.trufflesqueak.model.BlockClosureObject;
 import de.hpi.swa.trufflesqueak.model.CompiledCodeObject;
 import de.hpi.swa.trufflesqueak.model.ContextObject;
-import de.hpi.swa.trufflesqueak.model.NilObject;
 import de.hpi.swa.trufflesqueak.model.layout.ObjectLayouts.ASSOCIATION;
 import de.hpi.swa.trufflesqueak.nodes.AbstractNode;
 import de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectAt0NodeGen;
@@ -424,9 +423,9 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         }
     }
 
-    protected final void sendMustBeBooleanInInterpreter(final VirtualFrame frame, final int pc, final Object stackValue) {
+    protected final void sendMustBeBooleanInInterpreter(final VirtualFrame frame, final int pc, final int sp, final Object stackValue) {
         CompilerDirectives.transferToInterpreter();
-        FrameAccess.setInstructionPointer(frame, pc);
+        externalizePCAndSP(frame, pc, sp);
         final SqueakImageContext image = getContext();
         image.mustBeBooleanSelector.executeAsSymbolSlow(image, frame, stackValue);
         throw SqueakException.create("Should not be reached");
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -370,10 +370,9 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
     }
 
     protected static final Object pop(final VirtualFrame frame, final int sp) {
+        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
         final int slotIndex = FrameAccess.getStackStart() + sp;
-        final Object result = frame.getObjectStatic(slotIndex);
-        frame.setObjectStatic(slotIndex, NilObject.SINGLETON);
-        return result;
+        return frame.getObjectStatic(slotIndex);
     }
 
     protected static final Object popReceiver(final VirtualFrame frame, final int sp) {
@@ -386,12 +385,11 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
 
     @ExplodeLoop
     private static Object[] popNExploded(final VirtualFrame frame, final int sp, final int numPop) {
+        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
         final int firstSlotIndex = FrameAccess.getStackStart() + sp - numPop;
         final Object[] stackValues = new Object[numPop];
         for (int i = 0; i < numPop; i++) {
-            final int slotIndex = firstSlotIndex + i;
-            stackValues[i] = frame.getObjectStatic(slotIndex);
-            frame.setObjectStatic(slotIndex, NilObject.SINGLETON);
+            stackValues[i] = frame.getObjectStatic(firstSlotIndex + i);
         }
         return stackValues;
     }
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsInLoopNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsInLoopNode.java
@@ -47,7 +47,7 @@ public final class CheckForInterruptsInLoopNode extends AbstractNode {
         return SINGLETON;
     }
 
-    public void execute(final VirtualFrame frame, final int pc) {
+    public void execute(final VirtualFrame frame, final int pc, final int sp) {
         final SqueakImageContext image = getContext();
         final CheckForInterruptsState istate = image.interrupt;
         if (istate.shouldSkip()) {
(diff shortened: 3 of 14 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -35,13 +35,20 @@ public class ContextPrimitives extends AbstractPrimitiveFactoryHolder {
...
-        @Specialization(guards = {"0 <= newStackPointer", "newStackPointer <= receiver.size()"})
-        protected static final ContextObject store(final ContextObject receiver, final long newStackPointer) {
-            /*
-             * Not need to "nil any newly accessible cells" as cells are always nil-initialized and
-             * their values are cleared (overwritten with nil) on stack pop.
-             */
-            receiver.setStackPointer((int) newStackPointer);
+        @Specialization(guards = {"0 <= newStackPointerLong", "newStackPointerLong <= receiver.size()"})
+        protected static final ContextObject store(final ContextObject receiver, final long newStackPointerLong) {
+            final int oldStackPointer = receiver.getStackPointer();
```
fniephaus: 🤖 This now calls `receiver.getStackPointer()` before `receiver.setStackPointer(...)`. For contexts without a materialized Truffle frame yet, `getStackPointer()` goes through `getTruffleFrame()`, while `setStackPointer()` is the path that would have materialized the frame. So primitive 76 regresses from "materialize and update" to "fail before materialization" for non-materialized contexts. 🤖

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -172,13 +179,20 @@ protected static final Object doContextObject(final ContextObject receiver, fina
...
+         * asking for the frameSize of its method. Any fields above the stack pointer (stackp) are
+         * truly invisible even (and especially!) to the garbage collector. Any store into stackp
+         * other than by the primitive method stackp: is potentially fatal."
+         */
         @Specialization(guards = "receiver.hasTruffleFrame()")
         protected static final long doSize(final ContextObject receiver) {
             return receiver.getStackPointer();
         }
 
         @Specialization(guards = "!receiver.hasTruffleFrame()")
```
fniephaus, line 189: 🤖 Primitive 212 now has representation-dependent behavior: materialized contexts return `stackPointer`, but non-materialized ones still return `receiver.size() - receiver.instsize()`. That means `Context>>size` changes depending on whether the context has been materialized, which does not line up with the new "fields above stackp are invisible" rule. I think this branch still needs a single defined semantic plus a regression test. 🤖

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -35,13 +35,20 @@ public class ContextPrimitives extends AbstractPrimitiveFactoryHolder {
...
+        protected static final ContextObject store(final ContextObject receiver, final long newStackPointerLong) {
+            final int oldStackPointer = receiver.getStackPointer();
+            final int newStackPointer = (int) newStackPointerLong;
+
+            /* Nil any newly accessible or newly hidden stack slots */
+            final int start = Math.min(oldStackPointer, newStackPointer);
+            final int end = Math.max(oldStackPointer, newStackPointer);
+
+            for (int i = start; i < end; i++) {
+                receiver.atTempPut(i, NilObject.SINGLETON);
```
fniephaus, line 44: 🤖 Thanks for the contribution. This now clears every slot between the old and new stack pointer via `atTempPut`, and `atTempPut` also rewrites `frame.getArguments()` for argument/copied-value indices. So shrinking `stackp` below the argument/copied-value region no longer just hides those entries, it destroys them permanently if the stack pointer is raised again later. If that is the intended OpenSmalltalk-compatible behavior, I think this needs an explicit regression test; otherwise this looks broader than “lazy nilling of dead stack slots.”

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
```diff
@@ -36,6 +37,9 @@ public ResumeContextRootNode(final ResumeContextRootNode original) {
     @Override
     public Object execute(final VirtualFrame frame) {
         assert !activeContext.isDead() : "Terminated contexts cannot be resumed";
+        // Make it so ObjectGraphUtils can find the active context when executing
+        // strictly within the active context.
+        FrameAccess.setContext(frame, activeContext);
```
fniephaus, line 42: 🤖 Thanks for the careful fix here. Making the wrapper root frame point at `activeContext` looks right, but it is subtle because the interpreter still runs on `activeContext.getTruffleFrame()` while object-graph traversal walks the active call frames. I would strongly prefer a regression test that exercises `allInstances`/GC/finalization while executing through a resumed context, because this kind of reachability bug is easy to reintroduce later.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/FrameAccess.java
```diff
@@ -253,57 +248,119 @@ private static void clearStackSlots(final Frame frame) {
...
+            final int sp = getStackPointer(frame);
             final FrameDescriptor frameDescriptor = frame.getFrameDescriptor();
-            for (int slotIndex = SlotIndices.STACK_START; slotIndex < frameDescriptor.getNumberOfSlots(); slotIndex++) {
+            final int slotCount = frameDescriptor.getNumberOfSlots();
+            final int slotLimit = Integer.min(SlotIndices.STACK_START + sp, slotCount);
+            for (int slotIndex = SlotIndices.STACK_START; slotIndex < slotLimit; slotIndex++) {
                 action.accept(frame.getObjectStatic(slotIndex));
             }
+            /* Nil unreachable stack entries. */
+            if (frameHandling == FrameHandling.SCRUB) {
```
fniephaus, line 328: 🤖 Thanks for working through the GC/rooting details here. The new `SCRUB` path clears ordinary stack slots in live frames, but auxiliary slots are still only traversed, not scrubbed, while dead-frame scrubbing clears them. If auxiliary slots can ever hold object references, `flushDeadReferences()` can still retain objects that are logically unreachable. I would either scrub/assert auxiliary slots here as well, or document and enforce that they never participate in object reachability.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -149,7 +156,7 @@ protected static final AbstractSqueakObject doNextHandler(final ContextObject re
     @GenerateNodeFactory
     @SqueakPrimitive(indices = 210)
     protected abstract static class PrimContextAtNode extends AbstractPrimitiveNode implements Primitive1WithFallback {
-        @Specialization(guards = {"index <= receiver.size()"})
+        @Specialization(guards = {"1 <= index", "index <= receiver.getStackPointer()"})
```
fniephaus, line 155: 🤖 The new guard on primitives 210/211 seems too strict. These primitives still dispatch through `ContextObjectReadNode` / `ContextObjectWriteNode`, which support the full context layout (`sender`, `pc`, `stackp`, `method`, `closure`, `receiver`, temps), but the guard now limits access to `1..stackPointer`. As a result, valid fixed slots like `method` or `closure` can become inaccessible whenever `stackPointer` is smaller than those indices. I think the live-stack restriction should only apply to temp slots, not to the fixed part of the context object.

fniephaus:
🤖 One follow-up concern that I could not attach inline because the file is outside the diff: the lazy-nilling transition looks incomplete in the FFI/plugin stack helpers. `InterpreterProxy.setStackPointer()` / `pop()` and the return-value pop in `PrimExternalCallNode` directly decrement the frame stack pointer without clearing the newly hidden slot(s). That leaves stale objects above SP on these paths until a later scrub, even though the PR already had to adjust several other primitives for exactly this semantic change. It would be worth checking whether plugin code relies on the old eager-nil behavior here as well.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: Fixed! The old stack pointer is now considered zero if the Context does not yet have a Truffle frame.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: Fixed! If the Context does not have a Truffle frame, its stack pointer should be considered to be zero, so `size` should return zero.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: In OSVM, it is implemented as follows:

```
primitiveStoreStackp
	"Atomic store into context stackPointer. 
	Also ensures that any newly accessible cells are initialized to nil "
	| ctxt newStackp stackp |
	ctxt := self stackValue: 1.
	newStackp := self stackIntegerValue: 0.
	self success: newStackp >= 0.
	self success: newStackp <= (LargeContextSlots - CtxtTempFrameStart).
	successFlag ifFalse: [^ self primitiveFail].
	stackp := self fetchStackPointerOf: ctxt.
	"Nil any newly accessible cells"
	stackp + 1 to: newStackp do:
		[:i | self storePointerUnchecked: i + CtxtTempFrameStart - 1 ofObject: ctxt withValue: nilObj].
	self storeStackPointerValue: newStackp inContext: ctxt.
	self pop: 1
```

OSVM permits losing the arguments and temporaries and is counting on the Smalltalk code being well behaved :)

The implementation in this PR does the same thing for increasing the stack pointer value. When decreasing the stack pointer value, the newly hidden slots are nilled to eliminate known dead slots, but the code could be changed to skip that nilling if you prefer (the nilling would happen just before the next garbage collect).

Whether that change is made or not, if you decrease the stack pointer and then increase it, OSVM will nil those hidden slots. Keeping the code as it is now causes the nilling to happen on the decrease operation where OSVM causes it to happen on the increase operation (assuming the Context is on the heap and not on the stack).

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: These primitives are special versions of `at:` and `at:put:` for Context receivers. Just like their counterparts for other kinds of objects, they access the indexed variables of the object and not the fixed variables. To access the fixed variables (like `method` or `closure`) you would use `instVarAt:` or `instVarAt:put:`.

While it is true that the previous implementation of these primiitves could use negative indexes to access the fixed fields of the Context, OSVM definitely does not allow that.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/FrameAccess.java
Sabine-Schnabeltier: Since we don't use the auxiliary slots for storing object references, I've eliminated the iterations over them in `FrameAccess`.

Sabine-Schnabeltier:
Regarding `InterpreterProxy` and `PrimExternalCallNode`, it seems that we could avoid Context/frame stack manipulation entirely by using an emulated stack initialized by `Object[] receiverAndArguments` within `InterpreterProxy`. 

I haven't researched it fully, but it seems that the visible result from doing an external call is the value on the top of the stack in the frame and not any of the other stack values. So the receiver and arguments could be passed to the `interpreterProxy` at the beginning of `doExternalCall()` and the return value could be obtained from the `interpreterProxy` after the external code has been executed.

This would protect the interpreter from a poorly written external function that did not balance push and pop operations and would also ensure that the stack would be in the correct state if the primitive fails.

Sabine-Schnabeltier:
I added commits to make sure the internal SP is externalized when sending `mustBeBoolean` and I removed the frame access from `PrimExternalCallNode`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
Sabine-Schnabeltier: The following method succeeds with the change highlighted above but fails if the change is removed:

```
testResumedContextReachability
    | ready resume finished results |
    ready := Semaphore new.
    resume := Semaphore new.
    finished := Semaphore new.
    
    [
        [
            | uniqueObject instances |
            "Create object locally, only reachable from this Context."
            uniqueObject := MessageSend receiver: #TruffleSqueak selector: #reachabilityTest.
            
            "Force this Context to be run by ResumeContextRootNode when it resumes."
            ready signal.
            resume wait.
                        
            "Force GC while resumed."
            Smalltalk garbageCollect.
            
            "Gather the instances. If the Context isn't linked, uniqueObject will not be found among the instances."
            instances := MessageSend allInstances select: [ :each | 
                each selector == #reachabilityTest 
            ].
            
            "Send result to the main thread."
            results := { uniqueObject. instances }.
            
        ] ensure: [ 
            "Guarantee the main thread is woken up, even if an exception occurs"
            finished signal 
        ].
    ] fork.
    
    "Wait for the fork to initialize and suspend"
    ready wait.
    
    "Resume the fork, triggering ResumeContextRootNode"
    resume signal.
    
    "Wait for the fork to complete GC and data gathering"
    finished wait.
    
    self assert: results notNil description: 'Forked process failed to gather results'.
    self assert: results second size = 1.
    self assert: results second first == results first.
```

src/image-tests/runCuisTests.st
```diff
@@ -19,7 +19,6 @@ CompiledMethod preferredBytecodeSetEncoderClass == EncoderForSistaV1
 nonTerminatingTestCases := OrderedCollection new.
 {
     #ProcessTest -> #(#testResumeWithEnsureAfterBCR).
-    #TestValueWithinFix -> TestValueWithinFix allTestSelectors.
```
fniephaus, line 22: I had to disable this test when upgrading to GraalVM 25.1. I suspect there is some compiler bug or bug fix that is now triggering some sort of infinite loop, such as a Smalltalk while true loop that never checks for interrupts or so. Feel free to look into it, but it's probably unrelated to your changes.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
```diff
@@ -386,12 +385,11 @@ protected static final Object[] popN(final VirtualFrame frame, final int sp, fin
 
     @ExplodeLoop
     private static Object[] popNExploded(final VirtualFrame frame, final int sp, final int numPop) {
+        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
```
fniephaus, line 388: I think we can drop such comments. They will eventually be outdated. If you want to keep them around, turn them into javadoc and properly link `FrameAccess.iterateStackObjects()` so that we keep better track of this.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
```diff
@@ -370,10 +370,9 @@ protected static final void push(final VirtualFrame frame, final int sp, final O
     }
 
     protected static final Object pop(final VirtualFrame frame, final int sp) {
+        /* Unreachable stack entries are lazily nilled in FrameAccess.iterateStackObjects */
```
fniephaus, line 373: See https://github.com/hpi-swa/trufflesqueak/pull/250/changes/aa619c03bc58803edc0db26948f6e8f5a212964a#r3069683621

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsInLoopNode.java
```diff
@@ -56,6 +56,7 @@ public void execute(final VirtualFrame frame, final int pc) {
         /* Exclude interrupts case from compilation. */
         CompilerDirectives.transferToInterpreter();
         FrameAccess.setInstructionPointer(frame, pc);
+        FrameAccess.setStackPointer(frame, sp);
```
fniephaus, line 59: We should probably move `externalizePCAndSP()` from `AbstractInterpreterNode` into `FrameAccess`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
fniephaus: The primitive should stay as close as possible to OSVM behavior, regardless of what other tricks TruffleSqueak has up its sleeve ;)

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
```diff
@@ -172,13 +179,20 @@ protected static final Object doContextObject(final ContextObject receiver, fina
...
+         * other than by the primitive method stackp: is potentially fatal."
+         */
         @Specialization(guards = "receiver.hasTruffleFrame()")
         protected static final long doSize(final ContextObject receiver) {
             return receiver.getStackPointer();
         }
 
         @Specialization(guards = "!receiver.hasTruffleFrame()")
         protected static final long doSizeWithoutFrame(final ContextObject receiver) {
+            // ToDo: Is this correct? From the definition above, it should be zero...
```
fniephaus: Now that https://github.com/hpi-swa/trufflesqueak/pull/250/changes/aa619c03bc58803edc0db26948f6e8f5a212964a#r3065894598 is fixed, can this comment be removed?

fniephaus (changes requested):

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/plugins/ffi/InterpreterProxy.java
```diff
@@ -50,8 +50,9 @@ public final class InterpreterProxy {
     private static final int BaseHeaderSize = 8;
 
     private final SqueakImageContext context;
-    private MaterializedFrame frame;
     private int numReceiverAndArguments;
+    private Object[] receiverAndArguments;
```
fniephaus, line 52: Is there a reason we need `numReceiverAndArguments` now that there is `receiverAndArguments`? It seems it's possible that `receiverAndArguments` is `null`? When is that the case?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/plugins/ffi/InterpreterProxy.java
Sabine-Schnabeltier: It is null for this line in PrimExternalCallNode.lookupModuleLibrary():

                moduleInteropLibrary.invokeMember(library, "setInterpreter", context.getInterpreterProxy(0, null).getPointer());

I can remove the parameter since it is redundant.

fniephaus: And instead of null, you can pass EMPTY_ARRAY here.

Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ContextPrimitives.java
Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interrupts/CheckForInterruptsInLoopNode.java
Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
Sabine-Schnabeltier: Done!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
Sabine-Schnabeltier: Done!

src/image-tests/runCuisTests.st
Sabine-Schnabeltier: I reverted this commit

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ResumeContextRootNode.java
fniephaus: Added via https://github.com/hpi-swa/trufflesqueak/commit/653d85c53812d64ca0a9cd063be0098c5cd324d5

fniephaus (approved):
LGTM, thank you!

PR #251, opened 2026-04-15
Unify #cannotInterpret: and #doesNotUnderstand: dispatch fallbacks

Sabine-Schnabeltier:
Unify the handling of missing selectors (`#doesNotUnderstand:`) and missing method dictionaries (`#cannotInterpret:`) into a single fallback path.

`ClassObject#lookupInMethodDictSlow` now returns Java null on either structural failure. `resolveDispatchFailure(selector)` is used to walk the superclass chain and resolve the correct fallback method.

Added support for setting `methodDict` to nil or a MethodDictionary from Smalltalk.

Modifying `superclass` or `methodDict` now triggers a global method cache flush.

Flushing `#doesNotUnderstand:` or `#cannotInterpret:` now explicitly clears the entire global cache to prevent stale fallbacks.

`CompiledCodeObject`now invalidates its cached `primitiveNodeOrNull` if its headers or primitive-declaration bytecodes are mutated.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -1042,6 +1042,10 @@ public final class SqueakImageContext {
 
     /* Clear cache entries for selector (prim 119). */
     private void flushMethodCacheForSelector(final NativeObject selector) {
+        if (selector == doesNotUnderstand || selector == cannotInterpretSelector) {
+            flushMethodCache();
+            return;
+        }
         for (int i = 0; i < METHOD_CACHE_SIZE; i++) {
             if (methodCache[i].getSelector() == selector) {
                 methodCache[i].freeAndRelease();
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
@@ -65,7 +65,7 @@ public final class ClassObject extends AbstractSqueakObjectWithClassAndHash {
         instancesAreClasses = original.instancesAreClasses;
         superclass = original.superclass;
         assert superclass == null || superclass.assertNotForwarded();
-        methodDict = original.methodDict == null ? null : new VariablePointersObject(original.methodDict);
+        methodDict = original.methodDict != null ? new VariablePointersObject(original.methodDict) : null;
         format = original.format;
         // FIXME: should clone the pointers themselves, too
         pointers = original.pointers.clone();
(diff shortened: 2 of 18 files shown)


fniephaus (approved):
LGTM, thanks!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
```diff
@@ -423,8 +458,70 @@ private static Object lookupMethodInDictionary(final VariablePointersObject meth
...
+            }
+        } else {
+            // Search for #cannotInterpret: starting in the superclass of the missing method
+            // dictionary class.
+            final ClassObject startingClass = current.getSuperclassOrNull();
+            if (startingClass == null) {
+                throw SqueakException.create("Fatal dispatch error: hit nil method dictionary in class with no superclass while sending", originalSelector);
+            }
+
+            final CompiledCodeObject ciMethod = startingClass.lookupMethodInMethodDictSlow(image.cannotInterpretSelector);
```
fniephaus, line 499: 🤖 This only skips the first `nil` `methodDict`. If `startingClass` or any later superclass also has `methodDict := nil`, `lookupMethodInMethodDictSlow(#cannotInterpret:)` returns `null` again and we take the fatal not-found path instead of continuing to the next real ancestor. That seems reachable now that this PR explicitly allows `methodDict` to be `nil` from Smalltalk. I think the `#cannotInterpret:` search needs to keep walking past repeated structural failures rather than reusing the generic lookup helper here.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/CompiledCodeObject.java
```diff
@@ -727,8 +734,12 @@ public void initializeHeader(final long value) {
 
     public void setHeader(final long value) {
         final int oldNumLiterals = getNumLiterals();
+        final boolean oldHasPrimitive = hasPrimitive();
         internalHeader = CompiledCodeHeaderUtils.fromSmallIntegerValue(value);
         assert getNumLiterals() == oldNumLiterals;
+        if (oldHasPrimitive && !hasPrimitive()) {
```
fniephaus, line 740: 🤖 This reinitializes `primitiveNodeOrNull` only when a primitive flag is removed. The symmetric case matters too: if a method was previously non-primitive and had already cached `primitiveNodeOrNull = null`, then flipping the header to `hasPrimitive = true` leaves that stale `null` in place and the primitive path is never rediscovered. This probably needs invalidation whenever primitive-ness changes, not only on `true -> false`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
```diff
@@ -1042,6 +1042,10 @@ public void flushMethodCache() {
 
     /* Clear cache entries for selector (prim 119). */
     private void flushMethodCacheForSelector(final NativeObject selector) {
+        if (selector == doesNotUnderstand || selector == cannotInterpretSelector) {
+            flushMethodCache();
```
fniephaus, line 1046: 🤖 This protects the global method cache, but it still does not invalidate already-specialized fallback dispatch nodes. The `DispatchDirectMessageFallback*Node` variants hold a `DirectCallNode` to the resolved fallback method, and their assumptions are still derived from `lookupResult == null` rather than the fallback method `callTargetStable()`. After recompiling `#doesNotUnderstand:` or `#cannotInterpret:`, those inline caches can stay alive and keep calling the old fallback body even after this full flush.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/CompiledCodeObject.java
Sabine-Schnabeltier: Done, thanks!

PR #252, opened 2026-04-21
Enhance Dispatch Failure Robustness and Performance

Sabine-Schnabeltier:
### Description
This PR improves how the VM handles message dispatch failures. It integrates a DNU shortcut architecture into the dispatch paths and refines the handling of recursive failure scenarios.

### 1. Nested `#cannotInterpret:` Support
The VM now correctly handles "cascading" dispatch failures: If a message lookup fails (CI) and the subsequent attempt to send `#cannotInterpret:` also fails (CI), the VM now correctly nests these failures. `DispatchUtils` and related nodes now build nested `Message` objects, ensuring the original selector and arguments are preserved even across deep delegation chains.

### 2. Optimized DNU Shortcuts
Optional DNU shortcuts are now available for messages with 3 or fewer arguments. These shortcuts avoid the allocation of the arguments `Array` and the `Message` object by redirecting the failed dispatch to specialized, image-defined selectors.

* **Configuration:** The shortcuts are enabled by providing an array of symbols to the VM via the `primitiveSetDNUShortcutSelectors` primitive. The array maps message arity to a specific shortcut selector (index 1 for arity 0, index 2 for arity 1, etc.).
* **Form of the Shortcut:** When a dispatch failure occurs for an arity $N$ send, the VM checks if a shortcut is configured for that arity. If so, it dispatches a new message where the original arguments are followed by the original selector as the final argument.
    * *Example (Arity 2):* A failed send of `#at:put:` with args `[key, val]` is transformed into a send of the shortcut selector (e.g., `#dnu2:and:selector:`) with arguments `[key, val, #at:put:]`.
* **N-ary & Megamorphic Support:** This PR extends these shortcuts to the N-ary dispatch path and the indirect (megamorphic) path. By caching the arity as a compilation-final constant and using `@ExplodeLoop`, the JIT can unroll the shortcut frame creation into optimized machine code.

### 3. Cache Flushing & Stability
* **Absent Selector Assumptions:** Fallback nodes now track an image-global "absent selector" assumption. If a missing method is compiled into the image, the VM now correctly trips and invalidates these DNU nodes globally, preventing "stranded" DNU entries in the AST.
* **Stable Fallbacks:** The VM now tracks the stability of the fallback methods themselves (e.g., `#doesNotUnderstand:`), ensuring that if a developer redefines these error handlers, the AST updates immediately.

## Testing
* **Arity 0-3 Verification:** Verified that shortcuts for arities 0 through 3 correctly bypass standard DNU allocation and reach the expected Smalltalk methods.
* **Nested CI Verification:** Verified that corrupting multiple method dictionaries triggers the new nested `#cannotInterpret:` logic successfully, preserving the original message context.

[TruffleSqueakDispatchFallbackTests.st.zip](https://github.com/user-attachments/files/26916433/TruffleSqueakDispatchFallbackTests.st.zip)

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import java.util.logging.Level;
 
 import de.hpi.swa.trufflesqueak.util.DebugUtils;
+import org.graalvm.collections.EconomicMap;
 import org.graalvm.collections.UnmodifiableEconomicMap;
 
 import com.oracle.truffle.api.Assumption;
@@ -155,6 +156,7 @@ public final class SqueakImageContext {
     private static final int METHOD_CACHE_REPROBES = 4;
     private int methodCacheRandomish;
     @CompilationFinal(dimensions = 1) private final MethodCacheEntry[] methodCache = new MethodCacheEntry[METHOD_CACHE_SIZE];
+    private final EconomicMap<NativeObject, CyclicAssumption> absentSelectorAssumptions = EconomicMap.create();
 
     /* Interpreter state */
     private int primFailCode = -1;
@@ -234,8 +236,6 @@ public final class SqueakImageContext {
         }
         assert homePath != null && homePath.exists() : "Home directory does not exist: " + homePath;
         initializeMethodCache();
-
-        DebugUtils.registerContext(this);
     }
 
     public static SqueakImageContext get(final Node node) {
@@ -646,7 +646,39 @@ public final class SqueakImageContext {
         }
     }
 
+    @TruffleBoundary
+    public Assumption getAbsentSelectorAssumption(final NativeObject selector) {
+        CyclicAssumption absentAssumption = absentSelectorAssumptions.get(selector);
+        if (absentAssumption == null) {
+            absentAssumption = new CyclicAssumption("Absent selector globally: " + selector.asStringUnsafe());
+            absentSelectorAssumptions.put(selector, absentAssumption);
+        }
+        return absentAssumption.getAssumption();
+    }
+
+    @TruffleBoundary
+    private void invalidateAllAbsentSelectorAssumptions() {
+        for (final CyclicAssumption absentAssumption : absentSelectorAssumptions.getValues()) {
+            absentAssumption.invalidate("Fallback method (DNU/CI) shadowed or modified");
+        }
+        absentSelectorAssumptions.clear();
+    }
+
+    @TruffleBoundary
+    private void invalidateAbsentSelectorAssumption(final NativeObject selector) {
+        if (selector == doesNotUnderstand || selector == cannotInterpretSelector) {
+            invalidateAllAbsentSelectorAssumptions();
+            return;
+        }
+        final CyclicAssumption absentAssumption = absentSelectorAssumptions.get(selector);
+        if (absentAssumption != null) {
+            absentAssumption.invalidate("Absent selector flushed globally");
+            absentSelectorAssumptions.removeKey(selector);
+        }
+    }
+
     public void flushCachesForSelector(final NativeObject selector) {
+        invalidateAbsentSelectorAssumption(selector);
         flushCachesForSelectorInClassTable(selector);
         flushMethodCacheForSelector(selector);
     }
@@ -1045,10 +1077,6 @@ public final class SqueakImageContext {
 
     /* Clear cache entries for selector (prim 119). */
     private void flushMethodCacheForSelector(final NativeObject selector) {
-        if (selector == doesNotUnderstand || selector == cannotInterpretSelector) {
-            flushMethodCache();
-            return;
-        }
         for (int i = 0; i < METHOD_CACHE_SIZE; i++) {
             if (methodCache[i].getSelector() == selector) {
                 methodCache[i].freeAndRelease();
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -118,7 +118,7 @@ public final class SqueakImageContext {
     @CompilationFinal private ClassObject doubleByteArrayClass;
     @CompilationFinal private ClassObject wordArrayClass;
     @CompilationFinal private ClassObject doubleWordArrayClass;
-    public final NativeObject cannotInterpretSelector = new NativeObject(); // TODO: use selector
+    public final NativeObject cannotInterpretSelector = new NativeObject();
     public final ClassObject blockClosureClass = new ClassObject(this);
     @CompilationFinal private ClassObject fullBlockClosureClass;
     public final ClassObject largeNegativeIntegerClass = new ClassObject(this);
@@ -149,6 +149,11 @@ public final class SqueakImageContext {
                     CompiledCodeHeaderUtils.makeHeaderWord(true, 0, 0, 0, true, true),
                     ArrayUtils.EMPTY_ARRAY, compiledMethodClass);
     public final VirtualFrame externalSenderFrame = Truffle.getRuntime().createVirtualFrame(FrameAccess.newWith(NilObject.SINGLETON, null, NilObject.SINGLETON), dummyMethod.getFrameDescriptor());
+
+    // The maximum message arity that supports DNU shortcuts.
+    public static final int MAX_DNU_SHORTCUT_ARITY = 3;
+
+    @CompilationFinal(dimensions = 1) private NativeObject[] dnuShortcutSelectors = new NativeObject[0];
 
     /* Method Cache */
     private static final int METHOD_CACHE_SIZE = 2 << 12;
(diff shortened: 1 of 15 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/MethodCacheEntry.java
```diff
@@ -31,34 +30,34 @@ public Object getResult() {
         return result;
     }
 
-    public CompiledCodeObject getOrCreateFallbackMethod() {
-        if (fallbackMethod == null) {
-            fallbackMethod = createFallbackMethod();
+    public ClassObject.DispatchFailureResult getOrCreateDispatchFailureResult(final int arity) {
```
fniephaus, line 33: 🤖 `DispatchFailureResult` is now cached once per `(class, selector)`, but `resolveDispatchFailure` depends on arity. If the same missing selector is reached through
  `perform:withArguments:` with different argument counts, the first cached fallback can be reused for later sends and produce the wrong shortcut decision or wrong frame shape. This
  cache key needs to include arity, or the fallback result should be recomputed per send.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
```diff
@@ -1035,17 +1087,14 @@ public Object lookup(final ClassObject receiverClass, final NativeObject selecto
...
     /* Clear all cache entries (prim 89). */
     public void flushMethodCache() {
+        invalidateAllAbsentSelectorAssumptions();
         for (int i = 0; i < METHOD_CACHE_SIZE; i++) {
             methodCache[i].freeAndRelease();
         }
     }
 
     /* Clear cache entries for selector (prim 119). */
     private void flushMethodCacheForSelector(final NativeObject selector) {
```
fniephaus, line 1110: 🤖 This no longer clears cached fallback resolutions when `#doesNotUnderstand:` or `#cannotInterpret:` changes. `MethodCacheEntry` now stores the resolved fallback
  method separately from `result`, and this helper only drops entries whose message selector equals the flushed selector. A megamorphic send cached for some missing selector can
  therefore keep calling the old DNU or CI method after that fallback gets recompiled. I think the old global flush for fallback selectors is still needed here.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchUtils.java
```diff
@@ -65,6 +68,38 @@ static Assumption[] createAssumptions(final ClassObject startClass, final ClassO
...
+     * Fallback Method Stability: Tracks the `callTargetStable` of the resolved fallback method
+     * itself. This ensures the AST node is invalidated if the actual #doesNotUnderstand: or
+     * #cannotInterpret: method is later modified or recompiled.
+     *
+     * Absent Selector Stability: Tracks an image-global assumption that the specific failing
+     * selector does not exist. If a method for this missing selector is compiled, the VM's cache
+     * flush (primitive 119) will trip this assumption globally. This prevents "stranded DNU" nodes
+     * by forcing them to drop and re-resolve to the newly added method.
+     */
+    static Assumption[] getAssumptionsForMessageFallback(final Assumption[] classAssumptions, final NativeObject selector, final CompiledCodeObject fallbackMethod) {
```
fniephaus, line 84: 🤖 The fallback node assumptions only track absence of the original missing selector. If a shortcut selector is configured but missing when the node is created, the
  dispatch falls back to standard DNU, but compiling the shortcut method later will not invalidate that cached node. The optimization can stay stranded until a broader cache flush
  happens unless the configured shortcut selector is tracked here as well.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/MethodCacheEntry.java
Sabine-Schnabeltier: Thanks! Have added arity to the cache entry so that the proper `DispatchFailureResult` is returned if the arity changes.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
Sabine-Schnabeltier: Thanks! Rearranged where the selector check was made to handle this case.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchUtils.java
Sabine-Schnabeltier: Thanks! This should now be handled by `invalidateAllAbsentSelectorAssumptions()` when it is called via `flushCachesForSelector()` calling `flushMethodCache()`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -627,18 +699,33 @@ protected static final Object[] doMethod(final Node node, final AbstractSqueakOb
...
+                    return FrameAccess.newWith(sender, null, receiver, shortcutArgs);
+                }
+
+                final PointersObject message;
+                if (result.convention() == ClassObject.FallbackConvention.CANNOT_INTERPRET) {
+                    message = DispatchUtils.buildNestedMessage(createMessageNode, selector, result.fallbackSelector(), receiver, arguments, result.fallbackDepth());
+                } else {
+                    message = getContext(node).newMessage(writeNode, selector, receiverClass, arguments);
+                }
+                return FrameAccess.newMessageFallbackWith(sender, receiver, message);
```
fniephaus, line 747: 🤖 This N-ary indirect frame builder lost the object-as-method specialization that the fixed-arity builders still have. If `lookupResult` is a non-method object, `ResolveMethodNode` can return the object’s `run:with:in:` method, but this node now only handles compiled methods and `lookupResult == null`, so megamorphic N-ary object-as-method sends can end in `UnsupportedSpecializationException`. Please restore the old `doObjectAsMethod` specialization here, using the same `FrameAccess.newOAMWith(...)` shape as the fixed-arity paths.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -627,18 +699,33 @@ protected static final Object[] doMethod(final Node node, final AbstractSqueakOb
                 return FrameAccess.newWith(sender, null, receiver, arguments);
             }
 
-            @Specialization(guards = "lookupResult == null")
+            @Specialization(guards = {"lookupResult == null", "arguments.length == cachedArity"}, limit = "1")
```
fniephaus, line 705: 🤖 This specialization caches exactly one missing-send arity (`limit = "1"`) and there is no generic replacement specialization. The same indirect N-ary frame builder is used by megamorphic sends and `perform:withArguments:` paths, so after it specializes for one DNU arity, a later missing send with a different arity has no matching specialization and can fail with `UnsupportedSpecializationException`. Please keep the arity-specialized shortcut fast path, but add a replacing/generic fallback for other arities.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
```diff
@@ -478,31 +493,47 @@ public CompiledCodeObject resolveDispatchFailure(final NativeObject originalSele
...
+        // --- DNU Path (No nil method dictionaries) ---
         if (current == null) {
             assert originalSelector != image.doesNotUnderstand : "Fatal dispatch error: Recursive doesNotUnderstand:";
 
+            // Attempt the Shortcut DNU first
+            final NativeObject shortcutSelector = image.getDNUShortcutSelector(arity);
+            if (shortcutSelector != null) {
+                final CompiledCodeObject shortcutMethod = lookupMethodInMethodDictSlow(shortcutSelector);
+                if (shortcutMethod != null) {
+                    return new DispatchFailureResult(shortcutMethod, 1, FallbackConvention.SHORTCUT_DNU, shortcutSelector);
```
fniephaus: 🤖 This accepts a configured DNU shortcut method without checking that its argument count matches the shortcut frame shape (`arity + 1`). The primitive only validates that the configured entries are symbols/nil, so a bad image-side selector can make these direct fallback nodes call a method with the wrong frame layout. Please validate `shortcutMethod.getNumArgs() == arity + 1` before returning `SHORTCUT_DNU` and either ignore invalid shortcuts or fail deterministically.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
Sabine-Schnabeltier: Thanks for spotting this! The OAM case has been restored!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
Sabine-Schnabeltier: Done! Good catch!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ClassObject.java
Sabine-Schnabeltier: Added an arity confirmation check and also added a check to `primitiveSetDNUShortcutSelectors` to confirm that the argument contains either `Symbols` or `nil`.

fniephaus:
I haven’t forgotten about this and will try to integrate it for the next release!

Sabine-Schnabeltier:
Replaced by #299 and #307

PR #254, opened 2026-04-25
Merge extA and extB into extBA (supercedes #235)

Sabine-Schnabeltier:
This PR merges the `VirtualState` fields `extA` and `extB` (`int`s) into a single variable `extBA` (`long`). This will reduce the register pressure on the threaded interpreter, especially for Intel processors. This supercedes PR #235.

Changes:

- accessing `extA` and `extB` are now through helpers `getExtA()` and `getExtB()`
- replaced calls to `resetExtA()` and `resetExtB()` with `resetExtAB()`
- `processBytecode()` was modified to use `VirtualState` for tracking the extension register values
- decoding **EXT_A** and **EXT_B** byte codes and updating `extBA` is now handled in `VirtualState`
- decoding the **EXT_B** byte code now handles multibyte sequences properly

OSVM maintains a counter (`numExtB`) to determine whether an **EXT_B** byte code is the first in a series. Instead, we have been using `extB == 0` to indicate that the byte code is the first. At the moment, Squeak only emits single **EXT_B** byte codes, so this method will always work.

However, when the image starts using _sequences_ of **EXT_B** byte codes and the value being encoded is a positive integer with positions 7, 15 or 23 as the highest set bit, the decoded value will be interpreted as a negative integer. For example, the
emitted sequence for the value 0x8765 would be 0x00, 0x87, 0x65 and the second byte would be interpreted as a signed value in the current implementation.

If we assume that the encoder in the image will never try to encode the value 0 (which is safe since that is the default value without any emitted **EXT_B** byte codes), we can detect the case of the leading zero byte by setting the upper byte of `extB` and relying on the next byte to shift that byte out of the register.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/SqueakMiscellaneousTest.java
+++ b/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/SqueakMiscellaneousTest.java
@@ -18,6 +18,8 @@ import de.hpi.swa.trufflesqueak.model.CompiledCodeObject;
 import de.hpi.swa.trufflesqueak.model.FloatObject;
 import de.hpi.swa.trufflesqueak.model.NilObject;
 import de.hpi.swa.trufflesqueak.model.layout.ObjectLayouts.CONTEXT;
+import de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractDecoder.ShadowBlockParams;
+import de.hpi.swa.trufflesqueak.nodes.interpreter.DecoderSistaV1;
 import de.hpi.swa.trufflesqueak.nodes.interpreter.DecoderV3PlusClosures;
 import de.hpi.swa.trufflesqueak.util.UnsafeUtils;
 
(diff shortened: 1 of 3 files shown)


fniephaus (commented):
🤖 I found one consistency issue with the multi-byte EXT_B fix: the interpreter now handles the leading-zero sequence described in the PR, but DecoderSistaV1.decodeExtension() still uses the old EXT_B accumulation rule. For the PR example 0x00, 0x87, 0x65, the interpreter decodes 0x8765, while DecoderSistaV1 would still decode -30875.

That decoder path feeds stack-depth calculation, shadow block decoding, jump/block skipping, and other bytecode-analysis logic, so the fix should also update DecoderSistaV1 or share the EXT_B accumulation helper between the interpreter and decoder. A regression test for this exact sequence through a decoder path, such as an extended closure block size or extended jump, would make the behavior much harder to regress.

Sabine-Schnabeltier:
Great catch! I missed the mods to the decoder when i remade the PR.

The decoding of EXT_A and EXT_B are now helpers in InterpreterSistaV1Node. I also added a miscellaneous test for the extension decoding logic.

fniephaus (approved):
LGTM

PR #257, opened 2026-05-04
Do not skip restarted Contexts on marked CompiledMethods

Sabine-Schnabeltier:
In `ExecuteTopLevelContextNode`, `returnTo()` and `commonNVReturn()` skip over primitive Contexts assuming that the returned value is really destined for the sender of the primitive.

The current test checks if the Context is not for a closure, has an assigned primitive, and has PC = 0. However, the test does not take into account that unwind-marked and exception-handler-marked methods have primitives (without implementations, so always failing) and whose PC could be reset to zero when an exception handler is retried.

This can cause the Cuis 5 ExceptionTests #testSimpleRetry and #testSimpleRetryUsing to fail randomly: when the exception handler is restarted on the Truffle frame stack, the restart works and the test passes; if not on the Truffle stack, the test will fail.

I am not certain, but this may also be the cause of random failures of StringTest #testFindSelector.

This PR refines `ContextObject` `isPrimitiveContext()` to account for these marked Contexts and rearranges the conditions for the fewest number of tests during normal execution.

This PR also changes the handling of the return-top-from-method bytecodes in both Interpreters to pass a decremented stack pointer to `handleReturn()` since those bytecodes are supposed to pop the stack and return the value to the sender. This only makes a difference for block returns that send either #aboutToReturn:through: or #cannotReturn:

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
@@ -580,7 +580,9 @@ public final class ContextObject extends AbstractSqueakObjectWithHash {
 
     // The context represents primitive call which needs to be skipped when unwinding call stack.
     public boolean isPrimitiveContext() {
-        return !hasClosure() && getCodeObject().hasPrimitive() && getInstructionPointerForBytecodeLoop() == 0;
+        return getInstructionPointerForBytecodeLoop() == 0 && //
+                        !hasClosure() && !isUnwindMarked() && !isExceptionHandlerMarked() && //
+                        getCodeObject().hasPrimitive();
     }
 
     @TruffleBoundary
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
@@ -578,13 +578,6 @@ public final class ContextObject extends AbstractSqueakObjectWithHash {
         senderOrFrameOrSize = frame;
     }
 
-    // The context represents primitive call which needs to be skipped when unwinding call stack.
-    public boolean isPrimitiveContext() {
-        return getInstructionPointerForBytecodeLoop() == 0 && //
-                        !hasClosure() && !isUnwindMarked() && !isExceptionHandlerMarked() && //
-                        getCodeObject().hasPrimitive();
-    }
-
     @TruffleBoundary
     public boolean pointsTo(final Object thang) {
         // TODO: make sure this works correctly
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/InterpreterSistaV1Node.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/InterpreterSistaV1Node.java
@@ -790,7 +790,7 @@ public final class InterpreterSistaV1Node extends AbstractInterpreterNode {
                     }
                     case BC.RETURN_TOP_FROM_METHOD: {
                         state.reportLoopCountOnReturn(this);
-                        returnValue = handleReturn(frame, pc, pc + 1, vstate.sp, top(frame, vstate.sp));
+                        returnValue = handleReturn(frame, pc, pc + 1, vstate.sp - 1, top(frame, vstate.sp));
                         pc = LOCAL_RETURN_PC;
                         break;
                     }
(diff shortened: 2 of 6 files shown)


fniephaus:
Good catches!

> I am not certain, but this may also be the cause of random failures of StringTest #testFindSelector.

It seems `testFindSelector` passed in all jobs (both Squeak and Cuis), so maybe you are right. Should we unmark it as flaky as part of this PR?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
```diff
@@ -580,7 +580,9 @@ public void setTruffleFrame(final MaterializedFrame frame) {
 
     // The context represents primitive call which needs to be skipped when unwinding call stack.
     public boolean isPrimitiveContext() {
```
fniephaus: If I remember correctly, this was needed because primitive nodes were called right before entering the bytecode loop (see [here](https://github.com/hpi-swa/trufflesqueak/blob/4a39a58bc0bab7a674c576f6d914e266648673d4/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/ExecuteBytecodeNode.java#L73-L84) for example). For some time, however, primitives are tried before performing the call (see [here](https://github.com/hpi-swa/trufflesqueak/blob/bb26c4d3e195ee147d3e0000c982c071d19c718e/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java#L275) for example).

Have you seen your version of `isPrimitiveContext()` ever return `true`, or can we constant-fold it in the code now because it would always return `false`? :)

Sabine-Schnabeltier: I think you are correct!

But, doesn't that mean that if a Context with a primitive is restarted, the primitive will not be retried?

fniephaus: The context would not exist had the primitive succeeded, so I think this is expected behavior.

Sabine-Schnabeltier: I was thinking of the case where the primitive fails, execution continues in the Smalltalk fallback code and during that execution, the Context is restarted.

fniephaus: Sure, I was about the same case. IIRC primitives are tried before a method is activated, so I would assume the primitive is not retried on restart but I'm sure you can find out. 🙂

I think BitBlt's copyBits decompresses/unhibernates forms in the fallback code and then calls itself again (copyBitsAgain or so). Instead of the recursive call, you could try restarting the context. This might crash your image though. 😅

Sabine-Schnabeltier: > Sure, I was about the same case. IIRC primitives are tried before a method is activated, so I would assume the primitive is not retried on restart but I'm sure you can find out. 🙂

You are correct! The primitive is encoded as a CALL_PRIMITIVE bytecode at the very start of a method. Both interpreters simply skip over that bytecode (optionally storing the primitive failure result).

fniephaus: Even in the old version where the primitive was called right before the bytecode loop is entered in the same node, restarting a context wouldn’t have retried the primitive because it was handled outside the loop ;)

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/InterpreterSistaV1Node.java
```diff
@@ -790,7 +790,7 @@ public Object execute(final VirtualFrame frame_, final int startPC, final int st
                     }
                     case BC.RETURN_TOP_FROM_METHOD: {
                         state.reportLoopCountOnReturn(this);
-                        returnValue = handleReturn(frame, pc, pc + 1, vstate.sp, top(frame, vstate.sp));
+                        returnValue = handleReturn(frame, pc, pc + 1, vstate.sp - 1, top(frame, vstate.sp));
```
fniephaus, line 793: Why do we need to a decremented sp in handleReturn?

fniephaus (approved):
LGTM

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/InterpreterSistaV1Node.java
fniephaus: Oh, I missed this from the PR description:

> This PR also changes the handling of the return-top-from-method bytecodes in both Interpreters to pass a decremented stack pointer to handleReturn() since those bytecodes are supposed to pop the stack and return the value to the sender. This only makes a difference for block returns that send either #aboutToReturn:through: or #cannotReturn:

PR #258, opened 2026-05-05
Remove internalizePC()

Sabine-Schnabeltier:
In order to avoid a check for an updated PC after every message send, in this PR we force a flush of the Truffle execution stack; the updated PC will be used when the Context resumes execution.

It accomplishes this by checking bytecodes that write to receiver instance variables: when writing to the second instance variable (the program counter) of a Context object, it checks to see if the Context is on the Truffle frame stack. If the Context is found on the stack, a `ProcessSwitch` is raised to flush the Truffle stack and cause execution to resume with the updated program counter.

As a side effect of this PR, stack pointer resets caused by `Context>>privRefresh` will be properly picked up if the Context is on the Truffle stack, rather than being ignored as they are now.

This PR is based on changes in PR #257

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/ContextObject.java
@@ -8,6 +8,8 @@ package de.hpi.swa.trufflesqueak.model;
 
 import java.util.Arrays;
 
+import com.oracle.truffle.api.frame.Frame;
+import com.oracle.truffle.api.frame.FrameInstance;
 import org.graalvm.collections.UnmodifiableEconomicMap;
 
 import com.oracle.truffle.api.CallTarget;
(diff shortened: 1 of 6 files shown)


Sabine-Schnabeltier:
Even though `iterateFrames()` is expensive, it is called only when modifying the second instance variable of a `ContextObject`. Every other way I could think of to tell if a Context is on the Truffle frame stack requires an action when either restarting or exiting every Context which would be much higher overhead.

fniephaus:
> Even though iterateFrames() is expensive, it is called only when modifying the second instance variable of a ContextObject. Every other way I could think of to tell if a Context is on the Truffle frame stack requires an action when either restarting or exiting every Context which would be much higher overhead.

Right. Nonetheless, we always allocate `ContextObject`s before activating a method, so we can walk the stack from thisContext or the sender. I've pushed a commit that makes the changes and simplifies a couple of things.

fniephaus:
All tests are passing. Please take another look and let me know what you think. I must admit that I like `externalizePCAndSP()` and `internalizePC()` because that is close to OSVM, but TS deviates a lot any (e.g., it uses Truffle frames for method activations) and maybe PC modifications do deserve to be handled in a special way.

What do you think?

Sabine-Schnabeltier:
I agree it's a good idea to avoid `iterateFrames()` where possible. However, walking the sender chain has a different issue because it only gives us a definitive answer in one direction. If we walk the chain all the way to nil without finding the Context-to-be-modified, we know it is definitely not on the Truffle stack. But if we do find it along the way, it only tells us the logical Smalltalk state—we still don't know if it is actually suspended in an interpreter node loop on the physical Truffle stack. Plus, walking the entire chain to nil just to rule it out is a lot of overhead.

As an alternative to walking the sender chain, this new commit adds uses `booleanDBit` to track whether a Context `couldBeOnTruffleStack`. The first time a PC mutation is encountered for an executing Context, we flush the Truffle stack with a ProcessSwitch and flip the flag to indicate that the Context is definitively not on the physical stack anymore. (We then reset the flag when the context is later resumed).

This gives us the exact physical state guarantee we need, without the overhead of walking either the Truffle frames or the Squeak sender chain.

Sabine-Schnabeltier:
Added a Context-not-on-Truffle-stack state caching bit and associated documentation.

fniephaus (approved):
lgtm, thank you!

PR #259, opened 2026-05-07
Rework cannotReturn: to fix exception unwinding on dead contexts

Sabine-Schnabeltier:
Normal returns are supposed to send `cannotReturn:` to the current `Context` if either the `sender` is `nil` or the `Context` is dead.

This PR modifies normal returns in 2 ways:

1. **If `sender` has been modified:** Check that the new `sender` is not nil and not dead before throwing a `NonVirtualReturn`. Otherwise, throw `CannotReturn`. 
2. **If `sender` has not been modified:** Return the result (as before). If the sender is nil or dead, the current `Context` must be running within a `ResumeContextRootNode`, so control returns to `ExecuteTopLevelContextNode.returnTo()`.  In `returnTo()`, if the `sender` is not nil but is dead, we restore the `sender` in the returning `Context` *before* triggering the cannot-return logic. This keeps the sender chain intact so exception handlers can traverse it. (Previously, the sender was not restored so unwind and exception blocks below the sender were not found during exception handling.)

Additional changes:

* **Simplified NLR and NVR handlers:** By checking that the `sender` can be returned to *before* throwing NLR or NVR, we can remove checks for sending `#cannotReturn:`.
* **Explicit Context Creation (`CreateCannotReturnContextNode`):** Replaced the `Dispatch1Node` dispatch for `#cannotReturn:` with a dedicated node that manually creates a `Context` for evaluating `#cannotReturn:`. This lets us use `ExecuteTopLevelContextNode.executeLoop()` to handle the execution of `#cannotReturn:` without needing to separately handle any of the exceptions that could arise during its execution.
* **Test Un-skipping:** Removed `ProcessTest>>#testResumeWithEnsureAfterBCR` from the `nonTerminatingTestCases` list in `runCuisTests.st` since it now passes successfully.

This PR is based on changes in PRs #257 and #258

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -271,22 +271,27 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         if (isBlock) {
             return handleBlockReturn(frame, currentPC, pc, sp, result);
         } else {
-            return handleNormalReturn(frame, currentPC, result);
+            return handleNormalReturn(frame, currentPC, pc, sp, result);
         }
     }
 
-    protected final Object handleReturnFromBlock(final VirtualFrame frame, final int currentPC, final Object result, final int loopCounter) {
+    protected final Object handleReturnFromBlock(final VirtualFrame frame, final int currentPC, final int pc, final int sp, final Object result, final int loopCounter) {
         if (loopCounter > 0) {
             LoopNode.reportLoopCount(this, loopCounter);
         }
-        return handleNormalReturn(frame, currentPC, result);
+        return handleNormalReturn(frame, currentPC, pc, sp, result);
     }
 
-    protected final Object handleNormalReturn(final VirtualFrame frame, final int currentPC, final Object result) {
+    protected final Object handleNormalReturn(final VirtualFrame frame, final int currentPC, final int pc, final int sp, final Object result) {
         final byte profile = getProfile(currentPC);
         if (FrameAccess.hasModifiedSender(frame)) {
             enter(currentPC, profile, BRANCH1);
-            throw new NonVirtualReturn(result, FrameAccess.getSender(frame));
+            final AbstractSqueakObject sender = FrameAccess.getSender(frame);
+            if (sender instanceof ContextObject ctx && !ctx.isDead()) {
+                throw new NonVirtualReturn(result, sender);
+            } else {
+                throw cannotReturn(frame, pc, sp, result);
+            }
         } else {
             enter(currentPC, profile, BRANCH2);
             FrameAccess.terminateFrame(frame);
@@ -343,7 +348,6 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
 
     private static CannotReturnToTarget cannotReturn(final VirtualFrame frame, final int pc, final int sp, final Object returnValue) {
         CompilerDirectives.transferToInterpreter();
-        LogUtils.SCHEDULING.info("sendCannotReturn");
         FrameAccess.externalizePCAndSP(frame, pc, sp);
         throw new CannotReturnToTarget(returnValue, GetOrCreateContextWithFrameNode.executeUncached(frame));
     }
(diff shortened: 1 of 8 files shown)


fniephaus:
> This PR is based on changes in PRs https://github.com/hpi-swa/trufflesqueak/pull/257 and https://github.com/hpi-swa/trufflesqueak/pull/258

I removed the commits for the other PRs. #257 is merged already, and I pushed changes to #258. It turned out the changes in this PR did not depend on the two PRs, so I can review and merge this independently.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
```diff
@@ -271,22 +271,27 @@ protected final Object handleReturn(final VirtualFrame frame, final int currentP
...
 
-    protected final Object handleNormalReturn(final VirtualFrame frame, final int currentPC, final Object result) {
+    protected final Object handleNormalReturn(final VirtualFrame frame, final int currentPC, final int pc, final int sp, final Object result) {
         final byte profile = getProfile(currentPC);
         if (FrameAccess.hasModifiedSender(frame)) {
             enter(currentPC, profile, BRANCH1);
-            throw new NonVirtualReturn(result, FrameAccess.getSender(frame));
+            final AbstractSqueakObject sender = FrameAccess.getSender(frame);
+            if (sender instanceof ContextObject ctx && !ctx.isDead()) {
+                throw new NonVirtualReturn(result, sender);
```
fniephaus, line 290: This is the common path, so there is little point in adding another branch profile. cannotReturn is already behind a `transferToInterpreter()`.

fniephaus (approved):
LGTM, thanks for fixing another test!

fniephaus:
Mh, something might be wrong with the changes in this PR. The image seems to lock up on JVM on macOS:

- Run 1: https://github.com/hpi-swa/trufflesqueak/actions/runs/25601700258/job/75156620159?pr=259
- Run 2: https://github.com/hpi-swa/trufflesqueak/actions/runs/25601700258/job/75160037583?pr=259

Not sure how easy it is to reproduce this locally.

fniephaus:
🤖 **Potential correctness issue:** Non-local returns can still resume a dead target.

I think there is one remaining case where the new “validate before throwing NLR/NVR” model does not fully cover the target. `handleBlockReturn()` currently gates the NLR path on `homeContext.canBeReturnedTo()`, but `NonLocalReturn` does not return to `homeContext` itself. Its constructor stores `homeContext.getFrameSender()` as the target, and `commonNLReturn()` now assumes that target has already been validated and blindly pushes the return value into it.

That means a block whose home context is alive, but whose home sender is already dead, can still throw an NLR that resumes a dead context. This is especially risky because the PR removed the old top-level fallback check in `commonNLReturn()`, so the dead sender is no longer converted into `cannotReturn:` at the point where the NLR reaches the top-level loop.

Suggested fix: before throwing `NonLocalReturn`, validate the actual return target (`homeContext.getFrameSender()`), not just `homeContext.canBeReturnedTo()`. If the sender is not a live `ContextObject`, fall through to `cannotReturn(frame, pc, sp, result)` instead. This would keep the simplification in `commonNLReturn()` while preserving the invariant that every NLR target is returnable.

Sabine-Schnabeltier:
> Mh, something might be wrong with the changes in this PR. The image seems to lock up on JVM on macOS:
> 
> * Run 1: https://github.com/hpi-swa/trufflesqueak/actions/runs/25601700258/job/75156620159?pr=259
> * Run 2: https://github.com/hpi-swa/trufflesqueak/actions/runs/25601700258/job/75160037583?pr=259
> 
> Not sure how easy it is to reproduce this locally.

I have been seeing those random hangs on macOS on other branches (including main) over the past month or so, so I don't think it is directly related to this PR.

I looked into some earlier hangs yesterday which was the source of the FilePlugin PR changes that you accepted. Perhaps we could print the clock time at the start of each test to see if CI is getting bogged down?

fniephaus:
> Perhaps we could print the clock time at the start of each test to see if CI is getting bogged down?

The raw build job log contains timestamps. What would be better is some sort of watchdog that dumps all processes, saves, and then quits the image should a test take longer to run than say 2min or so.

I restarted the failing job, let's see if it passes this time.

Sabine-Schnabeltier:
> 🤖 **Potential correctness issue:** Non-local returns can still resume a dead target.

Reworked `canBeReturnedTo()` and changed its name to `canReturnToSender()`: it now checks that the homeContext is not dead and that homeContext's sender is a Context and that that Context is not dead.

fniephaus (approved):
LGTM, thank you!

PR #260, opened 2026-05-08
Wait for PlatformEventLoop to start before trying to open SqueakDisplay

Sabine-Schnabeltier:
When starting, it is possible for the image to start running and request a Display refresh before the PlaformEventLoop has completed initializing SDL3. This PR adds a latch that PlatformEventLoop uses to indicate it has finished the initializations and that SqueakDisplay waits on before opening the window for Display.


fniephaus (approved):
lgtm, thanks!

PR #261, opened 2026-05-10
FilePlugin fixes

Sabine-Schnabeltier:
It is possible that the cause of random timeouts of tests on GitHub is due to the servers getting overwhelmed with stdout flushes, so I'm experimenting with removing the flush from stdout writes. Smalltalk-originated flushes are still available, so this should not cause any visible change to the output.

Along the way, I found an optimization (use buffer as same as write) and a bug (integer file writes were the wrong endianness, I think?) ;)


fniephaus (approved):
LGTM, thanks!

PR #262, opened 2026-05-12
Rework interrupts to be thread safe

Sabine-Schnabeltier:
**Possibly fixes 30-minute test hangs on macOS ARM64 CI runners.**

There are two concurrency bugs:

1. **Silent Thread Crashes:** Replaced the non-thread-safe `ArrayDeque` with `ConcurrentLinkedDeque` for `semaphoresToSignal`. This prevents concurrent modification exceptions from silently killing the background timer thread.
2. **JMM Visibility Deadlock:** Changed `shouldTrigger` to use `VarHandle` with opaque access. This prevents GraalVM's JIT from caching the boolean in a CPU register (which possibly blinded the main thread to interrupts), while avoiding the heavy performance penalty of a `volatile` memory fence.

Restored the **TestValueWithinFix** tests since they may have been failing because of the problem this PR fixes.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -429,8 +429,8 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         return (bytecode & 7) + 1;
     }
 
-    protected final CheckForInterruptsInLoopNode createCheckForInterruptsInLoopNode(final int pc, final int numBytecodes, final int offset) {
-        return CheckForInterruptsInLoopNode.createForLoop(data, pc, numBytecodes, offset);
+    protected final CheckForInterruptsInLoopNode createCheckForInterruptsInLoopNode(final int targetPC, final int jumpOpcodePC) {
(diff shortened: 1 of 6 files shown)


Sabine-Schnabeltier:
It has run successfully 2 times; I've started 2 more tests; the PR should run it one more time. We can see if they all pass :)

Sabine-Schnabeltier:
I think that Graal may have been hoisting the read of `shouldTrigger` out of the interpreter backward loops. That may explain why the **TestValueWithinFix** tests were failing since some of those tests are tight loops without message sends to force an interrupt check.

I think the "random" hangs may have been caused by the multiple-thread accesses of the `ArrayDeque`.

Sabine-Schnabeltier:
I had to add back the `testResumeWithEnsureAfterBCR` test because it sometimes fails. I think that PR #258 may fix it for good.

Sabine-Schnabeltier:
I tried just using the `ConcurrentLinkedDeque` change only and it timed out every time (3 trials).

I think the tests worked yesterday in most places but I got a couple of failures on `testResumeWithEnsureAfterBCR`. Tonight, I am retrying 6 times to see if they all work successfully.

Result: Succeeds 5 out of 6 tries. So there is still some sort of race condition. Will look into it tomorrow.

Sabine-Schnabeltier:
Converted to a draft while I work out what is going on.

Sabine-Schnabeltier:
Replaced by #263

PR #263, opened 2026-05-18
Rework interrupts to be thread safe

Sabine-Schnabeltier:
Fixes 30-minute test hangs on macOS ARM64 CI runners. Also fixes the Cuis `TestValueWithinFix` tests since they were failing due to `shouldTrigger` being hoisted out of the loop. Ran `Test TruffleSqueak` 27 times without failure.

Changes include:

- `shouldTrigger` as a simple boolean can be hoisted outside the interpreter loop preventing interrupt checks. Convert to opaque accesses to guarantee it is always checked.
- `semaphoresToSignal` is accessed by multiple threads. Converted to `ConcurrentLinkedQueue`.
- Simplified the interrupt poll thread since all interrupt sources except the timer immediately set `shouldTrigger`.
- Added a `catch` block to the `CheckForInterruptsThread` polling thread to print error cause and terminate rather than silently exiting.
- Added a watchdog, settable from the command line, to dump the interpreter state and exit if execution continues beyond the given time in minutes. Use `--watchdog-timeout=timeInMinutes`
- Added a 25 minute watchdog to GitHub CI tests.
- Modified `PlatformEventLoop` to use 16 ms timeout when running in CI, guaranteeing progress even when GitHub macOS servers aggressively sleep UI threads. The wait event timeout remains infinite for GUI.
- Regenerated SDL3 bindings to replace `SDL_WaitEvent` with `SDL_WaitEventTimeout`. Used @timfel README as a guide to re-apply the Windows compatibility fix. `jextract` reordered `SDL_DestroyCursor` adding noise to the diff.

Replaces PR #262

### DIFF
--- a/src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
+++ b/src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
@@ -31,7 +31,7 @@ import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_Quit;
 import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_SetAppMetadataProperty;
 import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_SetEventEnabled;
 import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_SetHint;
-import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_WaitEvent;
+import static de.hpi.swa.trufflesqueak.sdl3.bindings.SDL_h.SDL_WaitEventTimeout;
 
 import java.lang.foreign.Arena;
 import java.lang.foreign.MemorySegment;
@@ -146,9 +146,13 @@ public final class PlatformEventLoop {
             final MemorySegment eventBuffer = SDL_Event.allocateArray(EVENT_FETCH_BATCH_SIZE, arena);
             final MemorySegment firstEvent = eventBuffer.asSlice(0, eventSize);
 
+            // Determine timeout: 16ms for CI, -1 (infinite) for GUI
+            final boolean isCI = Boolean.parseBoolean(System.getenv("CI"));
+            final int timeoutMS = isCI ? 16 : -1;
+
             while (isRunning) {
                 // Sleep until an event (or wakeUp() ping) arrives
-                if (SDL_WaitEvent(firstEvent)) {
+                if (SDL_WaitEventTimeout(firstEvent, timeoutMS)) {
                     eventHandler.accept(firstEvent);
                     // Peep additional events from the queue into our buffer
                     int eventsRead;
--- a/src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
+++ b/src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
@@ -94,7 +94,7 @@ public final class PlatformEventLoop {
         }
     }
 
-    public static void run() {
+    public static void run(final int sdlPollTimeoutMilliseconds) {
         try {
             startLatch.await();
         } catch (InterruptedException e) {
@@ -146,9 +146,7 @@ public final class PlatformEventLoop {
             final MemorySegment eventBuffer = SDL_Event.allocateArray(EVENT_FETCH_BATCH_SIZE, arena);
             final MemorySegment firstEvent = eventBuffer.asSlice(0, eventSize);
 
-            // Determine timeout: 16ms for CI, -1 (infinite) for GUI
-            final boolean isCI = Boolean.parseBoolean(System.getenv("CI"));
-            final int timeoutMS = isCI ? 16 : -1;
+            final int timeoutMS = sdlPollTimeoutMilliseconds > 0 ? sdlPollTimeoutMilliseconds : -1;
 
             while (isRunning) {
                 // Sleep until an event (or wakeUp() ping) arrives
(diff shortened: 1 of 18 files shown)


src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
```diff
@@ -146,9 +146,13 @@ public static void run() {
             final MemorySegment eventBuffer = SDL_Event.allocateArray(EVENT_FETCH_BATCH_SIZE, arena);
             final MemorySegment firstEvent = eventBuffer.asSlice(0, eventSize);
 
+            // Determine timeout: 16ms for CI, -1 (infinite) for GUI
+            final boolean isCI = Boolean.parseBoolean(System.getenv("CI"));
```
fniephaus: Mh, it’s unfortunate behavior has to deviate on CI. Why exactly? Can we really not find something that works in both cases?

isHeadless is not enough here because we run tests in GUI on CI?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
```diff
@@ -233,6 +234,11 @@ public SqueakImageContext(final SqueakLanguage squeakLanguage, final SqueakLangu
         }
         assert homePath != null && homePath.exists() : "Home directory does not exist: " + homePath;
         initializeMethodCache();
+
+        final int timeoutMinutes = options.watchdogTimeoutMinutes();
+        if (timeoutMinutes > 0) {
+            DebugUtils.startWatchdog(this, timeoutMinutes);
```
fniephaus: In theory, multiple contexts can be created (we used to do that for testing). Since the watchdog terminates the process, maybe we should be a bit more careful?

fniephaus: The option only really makes sense when used through the launcher, so maybe we can move this into the launcher somehow. We only need to find a way to trigger the logic you added to the DebugUtils

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
Sabine-Schnabeltier: Right, the tests are currently being run headful. The basic problem is that the GitHub CI runners put the GUI thread at a very low priority and so the SDL3 event loop  stalls.

If you prefer, I could add a command line option specifying the event loop timeout, defaulting to zero (meaning infinite).

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
Sabine-Schnabeltier: Is it possible to locate the SqueakImageContext from the launcher? If so, we could have the timeout occur in the launcher, locate the SqueakImageContext, and, from there, call the state dumping code in DebugUtils.

fniephaus: The launcher doesn’t have access to SqueakImageContext. It’s using the polyglot package designed for embedding guest languages in Java. It seems the polyglot Context can be interrupted, and it can be closed. Maybe we can install a handler for such interrupts or close events, I’m not sure. Maybe something to discuss on the #truffle channel on Slack.

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/PlatformEventLoop.java
fniephaus: Yeah, I think I’d prefer such an option. Thanks!

Sabine-Schnabeltier: I added --sdl-poll-timeout to the launcher options and set it for the CI tests.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
Sabine-Schnabeltier: I reworked the --watchdog-timeout option to be a launcher (rather than language) option and added a watchdog helper class to bridge between the launcher and the interpreter.

fniephaus:
mh, this job seems to have deadlocked suspiciously shortly after running `TestValueWithinFix` on Cuis:
https://github.com/hpi-swa/trufflesqueak/actions/runs/27453017049/job/81152025623?pr=263#step:16:398

Sabine-Schnabeltier:
Thanks for pointing this out! I think I may know what it is: the CheckForInterruptsInLoopNode is too liberal in deciding to be a no-op, Will see if I can find a solution...

Sabine-Schnabeltier:
Out of curiousity, how did you see where it hung?

fniephaus:
> Out of curiosity, how did you see where it hung?

I only looked at the [build log](https://github.com/hpi-swa/trufflesqueak/actions/runs/27453017049/job/81152025623?pr=263#step:16:398):

image

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/WatchdogBridge.java
```diff
@@ -0,0 +1,21 @@
...
+
+public final class WatchdogBridge {
+    private static volatile Runnable timeoutAction;
+
+    public static void setAction(final Runnable dumper) {
+        timeoutAction = dumper;
+    }
+
+    public static void triggerAction() {
+        if (timeoutAction != null) {
```
fniephaus, line 17: Should this throw an error if the action is triggered but the Runnable wasn't set? Right now, it just passes silently.

Sabine-Schnabeltier: Yes, great idea! Fixed!

fniephaus (approved):
LGTM, thanks!

PR #265, opened 2026-05-19
show test durations in GitHub Action Summary

Sabine-Schnabeltier:
This PR adds two summary tables showing the duration of each of the tests (Squeak, Cuis 7.3, Cuis 7.5) against each of the operating systems for JVM and AOT. Also adds a note to indicate if there were warnings or diagnostics generated during the run on the platform.


.github/workflows/ci.yml
```diff
@@ -145,29 +145,54 @@ jobs:
...
+        START_TIME=$SECONDS
         trufflesqueak ${{ env.VM_ARGS }} images/test-64bit.image $(pwd)/src/image-tests/runSqueakTests.st
+        ELAPSED_TIME=$(($SECONDS - $START_TIME))
+        FORMATTED_TIME=$(printf "%02d:%02d" $((ELAPSED_TIME / 60)) $((ELAPSED_TIME % 60)))
+        echo "SQUEAK=$FORMATTED_TIME" >> test-times.env
       timeout-minutes: 30
       if: ${{ matrix.os != 'windows-2022' }}
+    - &upload_test_times
+      name: Upload Test Times
+      uses: actions/upload-artifact@v7
```
fniephaus: I very much like the feature, but I would prefer an implementation that is simple and hopefully less error prone when GHA is unstable: can we have the job at the end use the Github CLI to enumerate the previous jobs and gather the time it took to run the tests? Maybe the API even makes the tuning info per step available? Or maybe there’s a way to fetch the log and search for some token?

Sabine-Schnabeltier: I will take a look!

Sabine-Schnabeltier:
Since you can't read the logs until the action has finished, I had to move the test summary to a different action, triggered by the completion of the Test TruffleSqueak action. At the same time, I moved the table to be a commit comment, like the benchmarks.

Sabine-Schnabeltier:
Changed this to post results to both the commit and the PR as per Slack discussion. Here's an example of the result: https://github.com/Sabine-Schnabeltier/trufflesqueak/pull/2

.github/workflows/ci.yml
```diff
@@ -145,27 +145,43 @@ jobs:
       name: Run Cuis 7.3 tests on TruffleSqueak standalone
       run: |
         mx.trufflesqueak/utils.sh download-cuis-7-3-test-image
+        START_TIME=$SECONDS
         trufflesqueak ${{ env.VM_ARGS }} images/Cuis?.?-????.image -s src/image-tests/runCuisTests.st
+        ELAPSED_TIME=$(($SECONDS - $START_TIME))
+        FORMATTED_TIME=$(printf "%02d:%02d" $((ELAPSED_TIME / 60)) $((ELAPSED_TIME % 60)))
+        echo "::TEST_TIME_METRIC::${{ matrix.type }}::${{ matrix.os }}::CUIS_73::${FORMATTED_TIME}::"
```
fniephaus: Instead of writing the timings to the log and download all logs, you can pass info via step outputs: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

.github/workflows/test-summary.yml
```diff
@@ -0,0 +1,126 @@
+name: Generate Master Test Summary
```
fniephaus: A separate workflow is not necessary if you use: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

.github/workflows/ci.yml
Sabine-Schnabeltier: Changed!

.github/workflows/test-summary.yml
Sabine-Schnabeltier: Changed!

.github/workflows/ci.yml
```diff
@@ -398,4 +462,122 @@ jobs:
...
+          done
+          
+          if [ "$FOOTNOTE_REQUIRED" = true ]; then
+            echo "*† Indicates that GraalVM compilation warnings or failure diagnostics were generated on this platform environment during the run.*" >> summary.md
+          fi
+          
+          # Echo to step summary for logging visibility
+          cat summary.md >> $GITHUB_STEP_SUMMARY
+          
+          echo "Posting summary directly to commit ${{ github.sha }}..."
```
fniephaus: I'd prefer to guard this, so it does not hit the GitHub API when running on a fork:
```
if [[ "${{ github.event_name }}" == "push" ]]; then
```

.github/workflows/ci.yml
```diff
@@ -398,4 +462,122 @@ jobs:
...
+          gh api \
```
fniephaus: For consistency, can we use `curl` here, just like in the `Analyze benchmark results` step?

.github/workflows/ci.yml
```diff
@@ -398,4 +462,122 @@ jobs:
...
         retention-days: 5
       if: ${{ matrix.kind == 'peak' }}
-  
+  test-summary:
+    name: Generate Test Summary
+    needs: [ build-jvm-standalone, build-native-standalone ]
+    if: always()
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write # Needed to post to the commit
```
fniephaus: We run `run-benchmarks` with default permissions and it works just fine. Can we drop this?

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -398,4 +462,122 @@ jobs:
...
+      - name: Download GraalVM Compiler Dumps (if any)
+        uses: actions/download-artifact@v4
+        with:
+          pattern: graal_dumps-*
+          path: all-times
+        continue-on-error: true
+
+      - name: Build Table and Post to Commit
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
fniephaus: Not required if we use `curl`:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -398,4 +462,122 @@ jobs:
...
+      contents: write # Needed to post to the commit
+    steps:
+      - name: Download GraalVM Compiler Dumps (if any)
+        uses: actions/download-artifact@v4
+        with:
+          pattern: graal_dumps-*
+          path: all-times
+        continue-on-error: true
+
+      - name: Build Table and Post to Commit
```
fniephaus:
```suggestion
      - name: Build table and post to commit
```

Sabine-Schnabeltier: Done!

.github/workflows/ci.yml
Sabine-Schnabeltier: Done!

.github/workflows/ci.yml
Sabine-Schnabeltier: Done!

.github/workflows/ci.yml
Sabine-Schnabeltier: Done!

.github/workflows/ci.yml
Sabine-Schnabeltier: Done!

.github/workflows/ci.yml
```diff
@@ -398,4 +462,124 @@ jobs:
...
         retention-days: 5
       if: ${{ matrix.kind == 'peak' }}
-  
+  test-summary:
+    name: Generate Test Summary
+    needs: [ build-jvm-standalone, build-native-standalone ]
+    if: always()
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Download GraalVM Compiler Dumps (if any)
```
fniephaus, line 471: Instead of downloading compiler dumps only to check which ones exist, we could also do the check in the jobs and set another output value to true or false, but maybe that adds too many outputs.

fniephaus (approved):
LGTM, thanks!

PR #272, opened 2026-06-18
Added --download-image launcher option

kilian-kier:
This PR introduces the support for using one of the supported images as a launcher option in a non-interactive way


kilian-kier:
Should I keep the pharo-13 image or should I remove it for now until it is usable?

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/SqueakLanguageOptions.java
```diff
@@ -22,6 +22,9 @@ public final class SqueakLanguageOptions {
     public static final String IMAGE_PATH = "image-path";
     public static final String IMAGE_PATH_FLAG = "--" + IMAGE_PATH;
     public static final String IMAGE_PATH_HELP = "Path to image";
+    public static final String DOWNLOAD_IMAGE = "download-image";
+    public static final String DOWNLOAD_IMAGE_FLAG = "--" + DOWNLOAD_IMAGE;
+    public static final String DOWNLOAD_IMAGE_HELP = "Download and run a supported image";
```
fniephaus: How about something like:
```suggestion
    public static final String DOWNLOAD_IMAGE_HELP = "Download a specific default image";
```

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/SqueakLanguageConfig.java
```diff
@@ -18,11 +18,12 @@ public final class SqueakLanguageConfig {
     private static final String IMAGE_VERSION = "25.0.1"; // on release: `VERSION;`
 
     public static final String[][] SUPPORTED_IMAGES = {
```
fniephaus: It's probably time to turn this into an array of records, maybe `SupportedImage[]` with `id`, `name`, and `url`?

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/SqueakImageLocator.java
```diff
@@ -79,6 +80,22 @@ private static int askUserToChooseImage(final String[][] supportedImages, final
         return selection;
     }
 
+    static String[] findSupportedImage(final String[][] supportedImages, final String imageKey) {
+        for (final String[] supportedImage : supportedImages) {
+            if (supportedImage[0].equals(imageKey)) {
+                return supportedImage;
+            }
+        }
+        final StringBuilder availableKeys = new StringBuilder();
```
fniephaus, line 113: Should we allow the new option to accept URLs (starting with `https://`)?

src/de.hpi.swa.trufflesqueak.launcher/src/de/hpi/swa/trufflesqueak/launcher/TruffleSqueakLauncher.java
```diff
@@ -43,6 +43,7 @@ public final class TruffleSqueakLauncher extends AbstractLanguageLauncher {
     private boolean quiet;
     private String[] imageArguments = new String[0];
     private String imagePath;
+    private String downloadImageKey;
```
fniephaus, line 46: It's a value here, so maybe just:
```suggestion
    private String downloadImage;
```

fniephaus (changes requested):

fniephaus:
> Should I keep the pharo-13 image or should I remove it for now until it is usable?

Let's add it when it's usable. If you add `https://` support, we can experiment with it on CI before adding it to the list.

src/de.hpi.swa.trufflesqueak.launcher/src/de/hpi/swa/trufflesqueak/launcher/TruffleSqueakLauncher.java
kilian-kier: It holds the image key you pass to the option. It is not the URL/record/name/path etc. so I think "key" in the field name makes sense

src/de.hpi.swa.trufflesqueak.shared/src/de/hpi/swa/trufflesqueak/shared/SqueakImageLocator.java
```diff
@@ -79,6 +80,26 @@ private static int askUserToChooseImage(final String[][] supportedImages, final
...
+            if (i > 0) {
+                availableKeys.append(", ");
+            }
+            availableKeys.append(supportedImages[i].id());
+        }
+        throw new RuntimeException("Not an URL and unknown image key '" + imageKey + "'. Available keys: " + availableKeys);
+    }
+
+    static boolean isDirectDownloadUrl(final String imageKey) {
+        return imageKey.startsWith("https://");
```
fniephaus: For convenience, we should probably throw a useful error message if the imageKey starts with https:// but doesn't end with .zip

fniephaus (approved):

PR #275, opened 2026-06-26
Refactor dispatch, optimize extended sends, and enhance DNU/CI handling

Sabine-Schnabeltier:
This PR refactors the message dispatch architecture, optimizes extended sends in the interpreter, and improves dispatch failure handling (DNU/CI). The multi-tiered dispatch caching architecture is based on a sketch by Fabio Niephaus (found in [wip/mult-level-dispatch](https://github.com/hpi-swa/trufflesqueak/tree/wip/multi-level-dispatch)). These changes reduce object allocations, resolve cache starvation for reflective calls, and locally improve overall test execution times. This PR builds on the work in PR #252.

## 1. Dispatch Cache Architecture

* **Centralized Management:** Introduces `DispatchCacheManager` for arity 0-5 and N-ary dispatch nodes.
* **Two-Tiered Caching:** Separates caching into a fast tier (standard methods and isolated fallbacks) and a wide tier (class polymorphism), with a fallback to indirect execution.
* **Fallback Isolation:** Isolates DNU and Object-As-Method (OAM) fallbacks in the fast tier to prevent improper coalescing of proxy classes.
* **N-ary Separation:** Splits N-ary dispatch into `DispatchNaryNode` (unchecked, for statically guaranteed bytecode) and `DispatchPerformNaryNode` (enforces dynamic array length checks via `executeWithCheckedArguments` for `#perform:withArguments:`).

## 2. Interpreter & Primitives

* **Extended Sends:** Unrolls extended send bytecodes for arities 3-5 in `InterpreterV3PlusClosuresNode` and `InterpreterSistaV1Node`, eliminating `popN()` `Object[]` allocations.
* **Reflective Calls:** Updates `PrimPerform` and `PrimPerformWithArguments` primitives to use the new cache managers. This caches the selector independently from the receiver class, preventing cache starvation on highly polymorphic reflective calls.

## 3. Dispatch Failures (DNU & CI)

* **DNU Shortcuts:** Adds optional, allocation-free shortcuts for failed sends up to arity 3, configured via `primitiveSetDNUShortcutSelectors`. A failed send transforms into a new dispatch that appends the original selector as the final argument.
* **Nested CI:** Supports cascading dispatch failures by building nested `Message` objects when `#cannotInterpret:` itself fails, preserving the original message context.
* **Cache Stability:** Fallback nodes now track an image-global "absent selector" assumption and the stability of the fallback method itself. This ensures proper AST invalidation when missing methods are added or error handlers are modified.

## Testing

* Verified DNU shortcuts correctly route arities 0-3 without standard DNU allocation.
* Verified corrupting multiple method dictionaries correctly triggers nested `#cannotInterpret:` logic.


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
```diff
@@ -48,27 +47,77 @@
...
+            while (currentFast != null) {
+                if (currentFast.guardChainNode.execute(receiver)) {
+                    return currentFast.dispatchDirectNode.execute(frame, receiver);
+                }
+                currentFast = currentFast.next;
+            }
+
+            // TIER 2 & 3: Wide Execution (Class Polymorphism)
+            if (cache.headWide != null) {
+                final ClassObject receiverClass = SqueakObjectClassNode.executeUncached(receiver);
```
fniephaus: Methods like `executeUncached()` should in most cases not be on the fast path. You could add a `SqueakObjectClassNode` child node to `WideDispatchDataNode`.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
```diff
@@ -48,27 +47,77 @@
...
+            // Node creation handles method resolution, including DNU and OAM fallbacks.
+            final DispatchDirect0Node newDirectNode = DispatchDirect0Node.create(selector, receiverClass, true);
+
+            final DispatchDirect0Node executor = cache.specialize(receiver, lookupResult, newDirectNode);
+
+            if (executor != null) {
+                return executor.execute(frame, receiver);
+            } else {
+                this.reportPolymorphicSpecialize();
+                indirectNode = insert(DispatchSelector0NodeFactory.DispatchIndirect0NodeGen.create());
```
fniephaus: `cache` should be cleared here, otherwise we keep the old PICs around

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
```diff
@@ -345,9 +454,21 @@ protected static final Object[] doMethod(final Node node, final AbstractSqueakOb
...
+                final ClassObject.DispatchFailureResult result = getContext(node).findMethodCacheEntry(receiverClass, selector).getOrCreateDispatchFailureResult(0);
+
+                if (result.convention() == ClassObject.FallbackConvention.SHORTCUT_DNU) {
+                    return FrameAccess.newWith(sender, null, receiver, selector);
+                }
+
                 final Object[] arguments = ArrayUtils.EMPTY_ARRAY;
-                final PointersObject message = getContext(node).newMessage(writeNode, selector, receiverClass, arguments);
+                final PointersObject message;
+                if (result.convention() == ClassObject.FallbackConvention.CANNOT_INTERPRET) {
```
fniephaus: It probably makes sense to use an `InlinedConditionProfile` here and for the `SHORTCUT_DNU` check above to reduce overall code size. Otherwise, all three cases are always part of compiled code.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
```diff
@@ -345,9 +454,21 @@ protected static final Object[] doMethod(final Node node, final AbstractSqueakOb
...
+                    message = DispatchUtils.buildNestedMessage(createMessageNode, selector, result.fallbackSelector(), receiver, arguments, result.fallbackDepth());
+                } else {
+                    message = getContext(node).newMessage(writeNode, selector, receiverClass, arguments);
```
fniephaus: Do a single `getContext(node)` or use `@Bind final SqueakImageContext image` in the parameter list.

Sabine-Schnabeltier:
### Multi-Level Dispatch PR: Comparison Test Findings

I ran a few simple tests to see how the new multi-level dispatch behaves under different configurations. Posting the summary of our Slack discussion here for future reference. Baseline is https://github.com/Sabine-Schnabeltier/trufflesqueak/commit/166bab27f000cc4514a337f4ec2eea0619978159

**Tested Configurations:**
1. Disable direct dispatch https://github.com/Sabine-Schnabeltier/trufflesqueak/commit/c7d8a0cf616adb57e74b50476f6e0c6f113d92d7
2. Disable wide dispatch https://github.com/Sabine-Schnabeltier/trufflesqueak/commit/1f865eb2fa4d3bbb46056271141545a7ce6a6854
3. Limit class polymorphism to 4 (before promotion to wide) https://github.com/Sabine-Schnabeltier/trufflesqueak/commit/905723a0557e2ce5f6770caebb24d4b6b5bd96ae

---

### Results

**1. Disable Direct Dispatch**
* The Peak Performance run timed out in the middle of the `Havlak` benchmark. My estimate from the tests that did complete is that the tests run about 10x slower.
* The Interpreter Performance is slightly slower (~20 s) without direct dispatch.

**2. Disable Wide Dispatch**
* **AWFY Benchmarks:** Slightly faster without wide dispatch (both peak and interpreted).
* **Test Durations:** Slightly slower overall.
* **Takeaway:** The AWFY benchmarks are not highly polymorphic, which likely explains why they aren't taking full advantage of the wide dispatch. 

**3. Limit Class Polymorphism to 4 (before wide promotion)**
* **AWFY Benchmarks:** Peak performance is essentially unchanged (~1 s slower), but interpreted is slightly faster (~13 s).
* **Test Durations (Mixed):**
  * **AOT:** Consistently faster with the smaller width.
  * **JVM:** Faster on Ubuntu Intel; roughly the same on Ubuntu ARM and Windows; slower on macOS.

---

### Additional Notes & Caveats
* **Environment Reliability:** GitHub runners are not dedicated benchmark machines. Linux ARM seems to provide the most reliable numbers, but overall, these metrics should be used to spot broad trends rather than exact benchmarks. 
* **Compilation Metrics:** Moving forward, it makes sense to compare the compilation summary at the end of test runs to see exactly how much code is being compiled across these different configurations.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
Sabine-Schnabeltier: Good point, thanks!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
Sabine-Schnabeltier: Good idea!
I had something in there before, but it was crashing the JVM, so I pulled it. Came up with an alternative way which I think will work.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
Sabine-Schnabeltier: Done, thanks!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
Sabine-Schnabeltier: Thanks!

fniephaus:
I had another look at the reports

- latest commit: https://github.com/Sabine-Schnabeltier/trufflesqueak/commit/49af5327e52822ada311d8c4a55da856e58ee832#commitcomment-191472388
- base commit: https://github.com/hpi-swa/trufflesqueak/commit/6577b033dd471d96dbfe8bf3ab2c3ccdcd9e03c3#commitcomment-191381950

It seems that interpreter performance is slightly worse (from 11:08 to 11:18), which is acceptable. There don't seem to be any clear improvements anymore. I thought we noticed that some tests run significantly faster? The ones that run on ARM are even slower. Not quite sure why that's the case.

Sabine-Schnabeltier:
I asked Gemini to compare `main` against this PR version and its verdict was the multi-level dispatch results in a slight improvement (as you've already noted). For some reason it left off `DeltaBlue`, which had the most improvement.
<hr>
It turns out that for the vast majority of these benchmarks, the multi-level dispatch implementation had absolutely no impact on the compilation metrics. However, there are two tests where GraalVM's behavior did shift.

Here is the complete breakdown of the remaining data:

### Benchmarks with Compilation Differences

| Benchmark | Metric | `main` | `nestingCI-dispatchML` | Difference |
| :--- | :--- | :--- | :--- | :--- |
| **Json** | Total Compilations | 136 | 140 | +4 Compilations |
| | Successful | 135 | 138 | +3 Successes |
| | Temporary Bailouts | 1 | 2 | +1 Bailout |
| | Invalidated | 4 | 4 | Identical |
| **Richards**| Total Compilations | 138 | 139 | +1 Compilation |
| | Successful | 138 | 138 | Identical |
| | Temporary Bailouts | 0 | 1 | +1 Bailout |
| | Invalidated | 4 | 3 | -1 Invalidation |

*(Note on Richards: Total compilations and temporary bailouts increased by one, but invalidations decreased by one.)*

---

### Benchmarks with Identical Compilation Behavior

For the remaining 9 benchmarks in your logs, GraalVM's compilation queues, successes, bailouts, and invalidations were **exactly the same** across both branches. The new dispatch logic did not alter the engine's compilation heuristics for these specific workloads. 

Here are the exact metrics both branches achieved:

* **Havlak:** 231 Compilations (230 Successes, 1 Temp Bailout, 15 Invalidated)
* **List:** 23 Compilations (23 Successes, 0 Temp Bailouts, 0 Invalidated)
* **Mandelbrot:** 11 Compilations (11 Successes, 0 Temp Bailouts, 1 Invalidated)
* **NBody:** 42 Compilations (42 Successes, 0 Temp Bailouts, 2 Invalidated)
* **Queens:** 25 Compilations (24 Successes, 1 Temp Bailout, 0 Invalidated)
* **Permute:** 19 Compilations (19 Successes, 0 Temp Bailouts, 0 Invalidated)
* **Sieve:** 20 Compilations (20 Successes, 0 Temp Bailouts, 0 Invalidated)
* **Storage:** 20 Compilations (20 Successes, 0 Temp Bailouts, 0 Invalidated)
* **Towers:** 26 Compilations (26 Successes, 0 Temp Bailouts, 0 Invalidated)


### Final Verdict

When combining this with your earlier data, the multi-level dispatch implementation proves to be highly stable. In the places where compilation behavior changes, it leans toward a net positive (significantly reducing compilations in heavy workloads like DeltaBlue) while causing only minor compiler thrashing in a couple of edge cases (Json, Richards). Most importantly, it rarely increases code invalidations, meaning GraalVM is still optimizing the generated graphs correctly.

Sabine-Schnabeltier:
Out of curiosity, I rebased your original implementation in `wip/multi-level-dispatch` and asked Gemini to compare the three versions:

<hr>

Here is a detailed structural breakdown comparing the core engine compilation metrics—**Total Compilations**, **Successes**, **Temporary/Permanent Bailouts**, **Invalidations**, and **Queue activity (Queues/Dequeues)**—across the three branches. 

---

## � Overview of Global Patterns
* **Permanent Bailouts:** There are **zero** permanent bailouts across all 15 test groups in every single run. The JIT compiler never encountered a method it completely gave up on.
* **Micro-benchmarks:** Small, stable benchmarks (*List, NBody, Queens, Permute, Sieve, Storage, Towers*) show identical metric counts across `main`, `wip`, and `merge`. The implementation differences do not alter compilation characteristics for these linear workloads.
* **Heavy Benchmarks:** The core differences appear exclusively in the behavior of the heavier benchmarks (*CD, DeltaBlue, Havlak, Json, Richards, Bounce*).

---

## � Deep Dive into Heavy Benchmarks

### 1. Compilations and Successes
This represents the total attempts by the JIT infrastructure to compile hot methods and the number of those attempts that completed cleanly.

| Benchmark | Metric | `main` (Reference) | `wip` (Test) | `merge` (PR) |
| :--- | :--- | :---: | :---: | :---: |
| **Bounce** | Compilations / Success | 36 / 35  | 34 / 34 | 35 / 35 |
| **CD** | Compilations / Success | 190 / 186 | 189 / 187 | 188 / 185 |
| **DeltaBlue** | Compilations / Success | 230 / 227 | 224 / 221 | 223 / 221 |
| **Havlak** | Compilations / Success | 231 / 230 | 227 / 225 | 231 / 230 |
| **Json** | Compilations / Success | 136 / 135 | 140 / 137 | 140 / 138 |
| **Richards** | Compilations / Success | 138 / 138 | 138 / 138 | 139 / 138 |

> **Key Takeaway:** The `merge` branch slightly reduces the absolute number of compilation attempts for **CD** and **DeltaBlue** compared to `main`, but increases compilation activity for **Json** and **Richards**. 

---

### 2. Temporary Bailouts
Temporary bailouts occur when Graal cancels a compilation midway (e.g., due to an invalidated assumption or thread cancellation) but intends to retry later.

* **Bounce:** `main` suffered 1 temporary bailout, whereas `wip` and `merge` successfully eliminated it (0 bailouts).
* **CD:** `wip` reduced bailouts to 2 (down from 4 in `main`). However, `merge` climbed back up to 3.
* **DeltaBlue:** `merge` achieved the lowest count with 2 bailouts, beating both `main` (3) and `wip` (3).
* **Json:** `main` was the most stable with 1 bailout. `wip` spiked up to 3, and `merge` settled in the middle at 2.
* **Richards:** `main` and `wip` had 0 bailouts, while `merge` introduced 1 temporary bailout.

---

### 3. Invalidations
Invalidations track how many times optimized machine code had to be thrown away at runtime because an assumption made during compilation was violated.

* **DeltaBlue:** This benchmark exhibits the highest code churn. `main` had 49 invalidations; `wip` saw an increase to 50; `merge` slightly improved the baseline, dropping down to 48.
* **Richards:** `merge` showed minor improvement here, tracking at 3 invalidations compared to 4 in both `main` and `wip`.
* **CD & Havlak:** Replicate identical numbers across all three runs (11 invalidations for CD and 15 for Havlak).

---

### 4. Queue Pressure (Queues vs. Dequeues)
The relationship between how many tasks enter the JIT queue vs. how many are dropped or abandoned (`Dequeues`) explains why compilation stalls occur.

**Havlak (The Anomalous Run):**
  * `main`: 390 queued / 160 dequeued
  * `wip`: 411 queued / 186 dequeued
  * `merge`: 391 queued / 161 dequeued
  
  In the `wip` branch, the system flooded the compiler queue with tasks (+21 compared to `main`) and discarded a massive amount of them (186 dequeues). Interestingly, `merge` reverted this behavior almost entirely back to the `main` baseline.
  
**DeltaBlue:**
  * `main`: 311 queued / 85 dequeued
  * `wip`: 313 queued / 91 dequeued
  * `merge`: 309 queued / 89 dequeued 
  
  `merge` slightly decreased overall queue entry pressure for DeltaBlue compared to the other two branches.

Sabine-Schnabeltier:
FWIW:

<hr>

Here is a breakdown of how the Smalltalk Are We Fast Yet (AWFY) benchmarks perform regarding these specific structural properties:

1. More than 8 Classes Sharing the Same Implementation via Inheritance
No, it is highly unlikely to find this within the AWFY suite.

This scenario describes high target polymorphism, which occurs when multiple different receiver classes point to the exact same method target in the lookup cache (usually because they inherit it from a shared superclass).

The AWFY suite is intentionally designed to evaluate fundamental language abstractions—like objects, closures, and arrays—without relying on sprawling frameworks or massive class libraries. As a result, the macro benchmarks have very compact, shallow class hierarchies. For example:

* **DeltaBlue** (a constraint solver) has one of the larger object hierarchies, but it only possesses around 6 distinct Constraint subclasses.

* **Json** features only a small handful of AST node types to represent JSON values.

Because the architectures of these benchmarks are purposefully constrained, they do not possess the structural width required to have more than 8 distinct classes sharing an identical, inherited method implementation.

2. More Than 4 Different Implementations of a Method
Yes, this does happen in the suite.

When a single method call site encounters multiple distinct receiver types with different implementations, the lookup cache can become saturated; caches holding a large number of entries (typically more than 4 in many VM architectures) are classified as megamorphic. While highly polymorphic call sites are generally rare (often representing less than 1% of method calls), they exist by design in specific AWFY benchmarks:

* **The Richards Benchmark**: This benchmark simulates an operating system kernel and its task scheduler. The scheduler continuously delegates processing to tasks, which are implemented via 5 distinct object roles: IdleTask, WorkerTask, HandlerA, HandlerB, and DeviceTask. Because the scheduler triggers the same execution method across all 5 of these distinct task objects, the call site sees 5 different implementations. This pushes the call site past the standard polymorphism threshold into megamorphism.

Sabine-Schnabeltier:
I worked with Gemini to create some benchmarks for comparing the multi-level dispatch against the current dispatch in main. It measures class polymorphic dispatches (1-10 classes sharing the same `CompiledMethod`) and message polymorphic dispatches (4-6 classes with independent `CompiledMethod`s). (The Smalltalk source for the benchmark is attached and the raw results are pasted below the summary.)

For low-polymorphism scenarios (1-4 classes) and method-polymorphic call sites, execution times remain on par with the current main branch. As expected, the performance gains are concentrated in sites with higher class polymorphism: execution times for 5 to 8 classes are approximately three times faster, and call sites with 9 or more classes run about 2.5 times faster.

| Configuration | Current Main Branch | New Multilevel Cache | Performance Delta |
| :--- | :--- | :--- | :--- |
| **Classes 1-4** | ~214 - 301 ms | ~219 - 322 ms | On par |
| **Classes 5-8** | ~770 - 915 ms | ~240 - 278 ms | **~3x Faster** |
| **Classes 9-10** | ~778 - 848 ms | ~309 - 349 ms | **~2.5x Faster** |
| **Methods 4** | ~226 - 250 ms | ~238 - 241 ms | On par |
| **Methods 5-6** | ~1035 - 1103 ms | ~1066 - 1070 ms | On par |


**Current Dispatch Results**
```
======================================================================
                   TRUFFLESQUEAK DISPATCH BENCHMARK                   
======================================================================
Configuration                   | Min (Peak)| Mean      | Max         
----------------------------------------------------------------------
Case 1: Classes (1)             | 214 ms    | 310 ms    | 393 ms
Case 1: Classes (2)             | 265 ms    | 338 ms    | 429 ms
Case 1: Classes (3)             | 265 ms    | 368 ms    | 468 ms
Case 1: Classes (4)             | 235 ms    | 297 ms    | 362 ms
Case 1: Classes (5)             | 915 ms    | 1011 ms   | 1163 ms
Case 1: Classes (6)             | 818 ms    | 944 ms    | 1014 ms
Case 1: Classes (7)             | 770 ms    | 831 ms    | 922 ms
Case 1: Classes (8)             | 783 ms    | 873 ms    | 1063 ms
Case 1: Classes (9)  [WIDE]     | 778 ms    | 838 ms    | 929 ms
Case 1: Classes (10) [WIDE]     | 848 ms    | 898 ms    | 1009 ms
----------------------------------------------------------------------
Case 2: Methods (4)             | 250 ms    | 333 ms    | 411 ms
Case 2: Methods (5) [INDIRECT]  | 1046 ms   | 1099 ms   | 1216 ms
Case 2: Methods (6) [INDIRECT]  | 1035 ms   | 1123 ms   | 1267 ms
======================================================================
```

```
======================================================================
                   TRUFFLESQUEAK DISPATCH BENCHMARK                   
======================================================================
Configuration                   | Min (Peak)| Mean      | Max         
----------------------------------------------------------------------
Case 1: Classes (1)             | 221 ms    | 279 ms    | 386 ms
Case 1: Classes (2)             | 301 ms    | 365 ms    | 417 ms
Case 1: Classes (3)             | 264 ms    | 357 ms    | 466 ms
Case 1: Classes (4)             | 250 ms    | 315 ms    | 387 ms
Case 1: Classes (5)             | 873 ms    | 938 ms    | 1035 ms
Case 1: Classes (6)             | 860 ms    | 987 ms    | 1095 ms
Case 1: Classes (7)             | 782 ms    | 863 ms    | 1030 ms
Case 1: Classes (8)             | 804 ms    | 864 ms    | 1003 ms
Case 1: Classes (9)  [WIDE]     | 803 ms    | 863 ms    | 981 ms
Case 1: Classes (10) [WIDE]     | 808 ms    | 899 ms    | 1002 ms
----------------------------------------------------------------------
Case 2: Methods (4)             | 226 ms    | 261 ms    | 293 ms
Case 2: Methods (5) [INDIRECT]  | 1101 ms   | 1182 ms   | 1324 ms
Case 2: Methods (6) [INDIRECT]  | 1103 ms   | 1181 ms   | 1320 ms
======================================================================
```

**Multi-level Dispatch Results**

```
======================================================================
                   TRUFFLESQUEAK DISPATCH BENCHMARK                   
======================================================================
Configuration                   | Min (Peak)| Mean      | Max         
----------------------------------------------------------------------
Case 1: Classes (1)             | 219 ms    | 476 ms    | 899 ms
Case 1: Classes (2)             | 274 ms    | 349 ms    | 406 ms
Case 1: Classes (3)             | 271 ms    | 330 ms    | 462 ms
Case 1: Classes (4)             | 233 ms    | 250 ms    | 305 ms
Case 1: Classes (5)             | 268 ms    | 352 ms    | 485 ms
Case 1: Classes (6)             | 255 ms    | 276 ms    | 315 ms
Case 1: Classes (7)             | 240 ms    | 268 ms    | 309 ms
Case 1: Classes (8)             | 249 ms    | 315 ms    | 445 ms
Case 1: Classes (9)  [WIDE]     | 309 ms    | 320 ms    | 334 ms
Case 1: Classes (10) [WIDE]     | 322 ms    | 401 ms    | 510 ms
----------------------------------------------------------------------
Case 2: Methods (4)             | 241 ms    | 318 ms    | 461 ms
Case 2: Methods (5) [INDIRECT]  | 1066 ms   | 1181 ms   | 1282 ms
Case 2: Methods (6) [INDIRECT]  | 1070 ms   | 1130 ms   | 1313 ms
======================================================================
```


```
======================================================================
                   TRUFFLESQUEAK DISPATCH BENCHMARK                   
======================================================================
Configuration                   | Min (Peak)| Mean      | Max         
----------------------------------------------------------------------
Case 1: Classes (1)             | 322 ms    | 425 ms    | 572 ms
Case 1: Classes (2)             | 252 ms    | 406 ms    | 689 ms
Case 1: Classes (3)             | 262 ms    | 289 ms    | 332 ms
Case 1: Classes (4)             | 260 ms    | 345 ms    | 469 ms
Case 1: Classes (5)             | 251 ms    | 282 ms    | 332 ms
Case 1: Classes (6)             | 267 ms    | 356 ms    | 491 ms
Case 1: Classes (7)             | 241 ms    | 247 ms    | 254 ms
Case 1: Classes (8)             | 245 ms    | 270 ms    | 317 ms
Case 1: Classes (9)  [WIDE]     | 349 ms    | 445 ms    | 519 ms
Case 1: Classes (10) [WIDE]     | 313 ms    | 430 ms    | 550 ms
----------------------------------------------------------------------
Case 2: Methods (4)             | 238 ms    | 328 ms    | 448 ms
Case 2: Methods (5) [INDIRECT]  | 1068 ms   | 1225 ms   | 1279 ms
Case 2: Methods (6) [INDIRECT]  | 1066 ms   | 1159 ms   | 1294 ms
======================================================================
```


[dispatch_benchmark.st.zip](https://github.com/user-attachments/files/30158915/dispatch_benchmark.st.zip)

fniephaus:
🤖 Correctness issue in `InterpreterV3PlusClosuresNode.handleExtendedSend()`: the new extended-send path pushes `result` directly, so it no longer applies `followForwarded(...)`. Previously this path used `sendNary(...)`, which did follow forwarding pointers; Sista’s new path still does this correctly. This can leave forwarded objects on the stack for V3 extended sends.

Suggested fix:
```java
push(frame, sp++, followForwarded(currentPC, result));
```

fniephaus:
🤖 DNU shortcut state/assumption issue in `SqueakImageContext.setDNUShortcutSelectors(...)`: calling `setDNUShortcutSelectors(null)` invalidates `dnuShortcutsAbsent` and then sets `dnuShortcutSelectors = null`. That makes the assertion in `hasDNUShortcuts()` inconsistent (`dnuShortcutSelectors == null` but the absent assumption is invalid), and hot-path specializations guarded by `image.getDnuShortcutsAbsent()` cannot become valid again after shortcuts are toggled off.

Since dispatch is on the hot path, it may be worth using a `CyclicAssumption` or otherwise recreating/separating the enabled/disabled-state assumptions so the no-shortcuts fast path can recover.

fniephaus:
🤖 I reran DeltaBlue from AWFY on Oracle GraalVM 25.1.3 with multi-tiering enabled:

```bash
trufflesqueak --jvm --experimental-options \
  --engine.TraceCompilation --engine.Mode=default \
  --headless --quiet -- \
  --evaluate "AWFYHarness run: #('DeltaBlue' 200 1000)"
```

3 runs each, comparing `main` against this PR.

### Timing results

| Branch | all 200 avg | last 100 avg | last 50 avg |
|---|---:|---:|---:|
| `main` | 7,783 µs | 4,783 µs | 4,213 µs |
| PR 275 | 7,318 µs | 4,303 µs | 3,727 µs |

PR improvement:
- All iterations: ~6.0% faster
- Last 100: ~10.0% faster
- Last 50: ~11.5% faster

### Compilation trace summary

| Branch | compilations/run | Tier 1/run | Tier 2/run | deopts/run | code size/run |
|---|---:|---:|---:|---:|---:|
| `main` | 153.7 | 109.7 | 44.0 | 26.0 | 820,464 |
| PR 275 | 150.0 | 109.3 | 40.7 | 25.0 | 752,255 |

Observations:
- `--engine.Mode=default` enabled Tier 2 compilation.
- PR compiles slightly fewer methods overall and fewer Tier 2 methods.
- PR has slightly fewer deopts.
- PR generates ~8.3% less compiled code.
- Total compilation time from traces was lower on PR: ~3,088 ms/run vs ~3,708 ms/run.

Conclusion: with multi-tiering enabled, this PR does look like an enhancement for DeltaBlue on TruffleSqueak JVM mode: it is meaningfully faster after warmup and has slightly cleaner compilation behavior with less generated code and fewer Tier 2 compilations/deopts.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/AbstractDispatchNode.java
```diff
@@ -16,6 +26,226 @@ public abstract class AbstractDispatchNode extends AbstractNode {
...
+
+            GuardChainDataNode current = head;
+            int count = 1;
+            while (current.next != null) {
+                current = current.next;
+                count++;
+            }
+
+            if (count < CacheLimits.LOOKUP_CACHE_LIMIT) {
+                current.next = current.insert(new GuardChainDataNode(receiver, assumptions));
```
fniephaus: It's easy to make mistakes and cause memory leaks or weird compilation issues when using raw node operations like `insert()` etc. Can this really not be expressed using the DSL and `@Cached`?

fniephaus:
My last commit triggered an interesting new failure:

```

Exception in thread "TruffleSqueakVM-Thread" org.graalvm.launcher.Launcher$AbortException
Caused by: org.graalvm.polyglot.PolyglotException: java.lang.ClassCastException: Stack trace is imprecise, the top frames are missing and/or have wrong line numbers. To get precise stack traces, build the image with option -H:-ReduceImplicitExceptionStackTraceInformation
	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNodeGen$Uncached.executeLookup(SqueakObjectClassNodeGen.java:615)
	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNode.executeUncached(SqueakObjectClassNode.java:43)
	at de.hpi.swa.trufflesqueak.nodes.dispatch.DispatchSelector1Node$Dispatch1Node.executeAndSpecialize(DispatchSelector1Node.java:108)
	at de.hpi.swa.trufflesqueak.nodes.dispatch.DispatchSelector1Node$Dispatch1Node.execute(DispatchSelector1Node.java:104)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.dispatch(AbstractInterpreterNode.java:183)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.dispatchHandled(AbstractInterpreterNode.java:176)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.send(AbstractInterpreterNode.java:171)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.handleSend1(InterpreterSistaV1Node.java:2474)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.__stub_handleSend1(InterpreterSistaV1Node.java:1)
	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.execute(InterpreterSistaV1Node.java:605)
	at de.hpi.swa.trufflesqueak.nodes.ResumeContextRootNode.execute(ResumeContextRootNode.java:55)
	at <smalltalk>.CTX WideStringTest#testAsIntegerSignedUsingRandomNumbers @66ebf717(Unknown)
	at org.graalvm.polyglot.Value.execute(Value.java:1096)
	at de.hpi.swa.trufflesqueak.launcher.TruffleSqueakLauncher.execute(TruffleSqueakLauncher.java:267)
	at de.hpi.swa.trufflesqueak.launcher.TruffleSqueakLauncher.lambda$launch$0(TruffleSqueakLauncher.java:173)
	at java.lang.Thread.runWith(Thread.java:1487)
	at java.lang.Thread.run(Thread.java:1474)
	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:915)
	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:898)
```

https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/29942564725/job/89004561163#step:19:6355

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/AbstractDispatchNode.java
```diff
@@ -16,6 +26,226 @@ public abstract class AbstractDispatchNode extends AbstractNode {
...
+                                currentFast.standardMethodOrNull == targetMethod &&
+                                currentFast.dispatchDirectNode.getClass() == newDispatchNode.getClass()) {
+
+                    if (currentFast.guardChainNode.append(receiver, newDispatchNode.getAssumptions())) {
+                        return currentFast.dispatchDirectNode;
+                    } else {
+                        // Guard chain overflow: Transition directly to wide execution
+                        removeFastNode(currentFast, previousFast);
+
+                        final WideDispatchDataNode<T> newWide = new WideDispatchDataNode<>(targetMethod, currentFast.dispatchDirectNode);
```
fniephaus: What's a bit unfortunate is that when this transitions from "fast" to "wide" dispatch, we essentially loose the profiling info stored in the guard chains in the fast dispatch. So when this happens, the `SqueakObjectClassNode` of the `WideDispatchDataNode` might go through several rounds of respecialization until it covers more than what the corresponding guard chain used to cover.

I don't have a solution to this yet. It's just a potential problem. And for particularly hot methods, Truffle may trigger recompilation relatively early, which would mean compilation time is wasted.

Sabine-Schnabeltier:
> My last commit triggered an interesting new failure:
> 
> ```
> 
> Exception in thread "TruffleSqueakVM-Thread" org.graalvm.launcher.Launcher$AbortException
> Caused by: org.graalvm.polyglot.PolyglotException: java.lang.ClassCastException: Stack trace is imprecise, the top frames are missing and/or have wrong line numbers. To get precise stack traces, build the image with option -H:-ReduceImplicitExceptionStackTraceInformation
> 	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNodeGen$Uncached.executeLookup(SqueakObjectClassNodeGen.java:615)
> 	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNode.executeUncached(SqueakObjectClassNode.java:43)
> 	at de.hpi.swa.trufflesqueak.nodes.dispatch.DispatchSelector1Node$Dispatch1Node.executeAndSpecialize(DispatchSelector1Node.java:108)
> 	at de.hpi.swa.trufflesqueak.nodes.dispatch.DispatchSelector1Node$Dispatch1Node.execute(DispatchSelector1Node.java:104)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.dispatch(AbstractInterpreterNode.java:183)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.dispatchHandled(AbstractInterpreterNode.java:176)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.AbstractInterpreterNode.send(AbstractInterpreterNode.java:171)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.handleSend1(InterpreterSistaV1Node.java:2474)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.__stub_handleSend1(InterpreterSistaV1Node.java:1)
> 	at de.hpi.swa.trufflesqueak.nodes.interpreter.InterpreterSistaV1Node.execute(InterpreterSistaV1Node.java:605)
> 	at de.hpi.swa.trufflesqueak.nodes.ResumeContextRootNode.execute(ResumeContextRootNode.java:55)
> 	at <smalltalk>.CTX WideStringTest#testAsIntegerSignedUsingRandomNumbers @66ebf717(Unknown)
> 	at org.graalvm.polyglot.Value.execute(Value.java:1096)
> 	at de.hpi.swa.trufflesqueak.launcher.TruffleSqueakLauncher.execute(TruffleSqueakLauncher.java:267)
> 	at de.hpi.swa.trufflesqueak.launcher.TruffleSqueakLauncher.lambda$launch$0(TruffleSqueakLauncher.java:173)
> 	at java.lang.Thread.runWith(Thread.java:1487)
> 	at java.lang.Thread.run(Thread.java:1474)
> 	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:915)
> 	at com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:898)
> ```
> 
> https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/29942564725/job/89004561163#step:19:6355

OK, spotted the problem; will get a fix in shortly...

Sabine-Schnabeltier:
> My last commit triggered an interesting new failure:
> 
> ```
> 
> Exception in thread "TruffleSqueakVM-Thread" org.graalvm.launcher.Launcher$AbortException
> Caused by: org.graalvm.polyglot.PolyglotException: java.lang.ClassCastException: Stack trace is imprecise, the top frames are missing and/or have wrong line numbers. To get precise stack traces, build the image with option -H:-ReduceImplicitExceptionStackTraceInformation
> 	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNodeGen$Uncached.executeLookup(SqueakObjectClassNodeGen.java:615)
> 	at de.hpi.swa.trufflesqueak.nodes.accessing.SqueakObjectClassNode.executeUncached(SqueakObjectClassNode.java:43)
> ```
> 
> https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/29942564725/job/89004561163#step:19:6355

This seems to be a duplicate of #285, so I am not sure that this is new to the changes in this PR.

I am separately submitting a PR regarding an issue where non-directed super sends could dispatch to a stale superclass. That bug causes silent dispatch errors (routing to the wrong superclass hierarchy) and could potentially lead to cast exceptions depending on the nature of the #becomeForward:, but given the obscured stack trace, it's hard to say for sure if it is the root cause here.

Sabine-Schnabeltier:
> 🤖 DNU shortcut state/assumption issue in `SqueakImageContext.setDNUShortcutSelectors(...)`: calling `setDNUShortcutSelectors(null)` invalidates `dnuShortcutsAbsent` and then sets `dnuShortcutSelectors = null`. That makes the assertion in `hasDNUShortcuts()` inconsistent (`dnuShortcutSelectors == null` but the absent assumption is invalid), and hot-path specializations guarded by `image.getDnuShortcutsAbsent()` cannot become valid again after shortcuts are toggled off.
> 
> Since dispatch is on the hot path, it may be worth using a `CyclicAssumption` or otherwise recreating/separating the enabled/disabled-state assumptions so the no-shortcuts fast path can recover.

Thanks for catching this! I've reworked the assumption so that the fast path ought to recover when the DNU shortcuts are disabled by being set to nil.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/AbstractDispatchNode.java
Sabine-Schnabeltier: > It's easy to make mistakes and cause memory leaks or weird compilation issues when using raw node operations like `insert()` etc. Can this really not be expressed using the DSL and `@Cached`?

I've tried mocking up a version that uses the DSL and `@Cached`, but I don't see a way to have mixtures of fast and wide nodes: if any fast node needs to go wide, they all go wide.

I think the mix of fast and wide nodes may be important for execution, so I made a version of the code in the PR that measures the distribution of the nodes just to see if that is the case. I'll post an AI generated summary of the findings in the next comment.

Sabine-Schnabeltier: ## Key Analytical Findings

1. **Monomorphism Baseline:** Across all large-scale workloads (test suites and interactive sessions), between **93.8% and 97.5%** of all dispatch sites remain monomorphic (`[1 fast, 0 wide]`).
2. **Interactive UI Workload Dynamics:** Interactive execution (Morphic UI rendering, browser activity, and event dispatch) increases wide-tier participation to **2.26%**, compared to **< 0.80%** in static test suites. Fast-tier dual-receiver nodes (`[2 fast, 0 wide]`) also increase to **3.31%**.
3. **Coexistence of Tiers:** The presence of non-zero counts in mixed states (e.g., `[1 fast, 1 wide]`, `[2 fast, 1 wide]`, and `[3 fast, 1 wide]`) confirms that fallback methods (such as `#doesNotUnderstand:`) remain isolated in the fast tier while standard methods undergo promotion to the wide tier.
4. **Indirect Fallback Rate:** Across all non-microbenchmark runs, the percentage of dispatch nodes exceeding the cache budget and falling back to indirect dispatch remains bounded between **0.13% and 0.27%**.


<details>
<summary><b>Click to expand full benchmark report & analysis</b></summary>

<h2 data-heading="Executive Overview">Executive Overview</h2>
<p>The dispatch cache tracks nodes using a two-dimensional matrix <code>[f fast, w wide]</code>, where <span class="math math-inline">f</span> represents the number of fast-tier entries (<code>FastDispatchDataNode</code>) and <span class="math math-inline">w</span> represents the number of wide-tier entries (<code>WideDispatchDataNode</code>).</p>
<ul>
<li><strong>Fast Tier:</strong> Stores standard methods and fallback scenarios (e.g., <code>#doesNotUnderstand:</code>) up to a set guard limit.</li>
<li><strong>Wide Tier:</strong> Handles class polymorphism when standard methods exceed fast-tier guard limits.</li>
<li><strong>Indirect Nodes:</strong> Dispatch sites where total method capacity exceeds <code>CacheLimits.DISPATCH_CACHE_LIMIT</code>, forcing a transition to indirect execution.</li>
</ul>
<hr>
<h2 data-heading="Workload Summaries">Workload Summaries</h2>
<h3 data-heading="1. AWFY Microbenchmark Suite (15 Benchmarks)">1. AWFY Microbenchmark Suite (15 Benchmarks)</h3>
<ul>
<li><strong>Node Scale:</strong> Each individual benchmark run generated between ~1,000 and ~1,600 total cached dispatch nodes.</li>
<li><strong>Monomorphic Dispatches:</strong> Ranged between 95% and 98% across all runs (<code>[1 fast, 0 wide]</code> accounted for 987 nodes in <em>Bounce</em> up to 1,535 nodes in <em>DeltaBlue</em>).</li>
<li><strong>Wide Tier Utilization:</strong> Exactly 1 node occupied the <code>[0 fast, 1 wide]</code> state per benchmark run, with 0 nodes in higher wide-tier states.</li>
<li><strong>Indirect Fallbacks:</strong> Consistently 1 or 2 indirect nodes per benchmark run.</li>
</ul>
<hr>
<h3 data-heading="2. Full System Test Suites (Cuis 7.3, Cuis 7.5, Squeak)">2. Full System Test Suites (Cuis 7.3, Cuis 7.5, Squeak)</h3>
<h4 data-heading="**Cuis 7.3 Tests**"><strong>Cuis 7.3 Tests</strong></h4>
<ul>
<li>
<p><strong>Total Cached Nodes:</strong> 72,058</p>
</li>
<li>
<p><strong>Monomorphic (<code>[1 fast, 0 wide]</code>):</strong> 68,573 nodes (95.16%)</p>
</li>
<li>
<p><strong>Polymorphic Fast Tier (<code>[2..4 fast, 0 wide]</code>):</strong> 3,071 nodes (4.26%)</p>
</li>
<li>
<p><strong>Wide Tier Nodes (<span class="math math-inline">w > 0</span>):</strong> 514 nodes (0.71%)</p>
</li>
<li>
<p><code>[0 fast, 1 wide]</code>: 380 | <code>[0 fast, 2 wide]</code>: 3</p>
</li>
<li>
<p><code>[1 fast, 1 wide]</code>: 110 | <code>[2 fast, 1 wide]</code>: 12 | <code>[3 fast, 1 wide]</code>: 7</p>
</li>
<li>
<p><code>[2 fast, 2 wide]</code>: 2</p>
</li>
<li>
<p><strong>Indirect Nodes:</strong> 187 nodes (0.26%)</p>
</li>
</ul>
<h4 data-heading="**Cuis 7.5 Tests**"><strong>Cuis 7.5 Tests</strong></h4>
<ul>
<li>
<p><strong>Total Cached Nodes:</strong> 139,171</p>
</li>
<li>
<p><strong>Monomorphic (<code>[1 fast, 0 wide]</code>):</strong> 135,640 nodes (97.46%)</p>
</li>
<li>
<p><strong>Polymorphic Fast Tier (<code>[2..4 fast, 0 wide]</code>):</strong> 2,987 nodes (2.15%)</p>
</li>
<li>
<p><strong>Wide Tier Nodes (<span class="math math-inline">w > 0</span>):</strong> 544 nodes (0.39%)</p>
</li>
<li>
<p><code>[0 fast, 1 wide]</code>: 408 | <code>[0 fast, 2 wide]</code>: 3</p>
</li>
<li>
<p><code>[1 fast, 1 wide]</code>: 110 | <code>[1 fast, 2 wide]</code>: 1 | <code>[2 fast, 1 wide]</code>: 12</p>
</li>
<li>
<p><code>[2 fast, 2 wide]</code>: 1 | <code>[3 fast, 1 wide]</code>: 9</p>
</li>
<li>
<p><strong>Indirect Nodes:</strong> 183 nodes (0.13%)</p>
</li>
</ul>
<h4 data-heading="**Squeak Tests**"><strong>Squeak Tests</strong></h4>
<ul>
<li>
<p><strong>Total Cached Nodes:</strong> 228,027</p>
</li>
<li>
<p><strong>Monomorphic (<code>[1 fast, 0 wide]</code>):</strong> 220,344 nodes (96.63%)</p>
</li>
<li>
<p><strong>Polymorphic Fast Tier (<code>[2..4 fast, 0 wide]</code>):</strong> 5,849 nodes (2.56%)</p>
</li>
<li>
<p><strong>Wide Tier Nodes (<span class="math math-inline">w > 0</span>):</strong> 1,834 nodes (0.80%)</p>
</li>
<li>
<p><code>[0 fast, 1 wide]</code>: 1,444 | <code>[0 fast, 2 wide]</code>: 17</p>
</li>
<li>
<p><code>[1 fast, 1 wide]</code>: 264 | <code>[1 fast, 2 wide]</code>: 4 | <code>[2 fast, 1 wide]</code>: 70</p>
</li>
<li>
<p><code>[2 fast, 2 wide]</code>: 1 | <code>[3 fast, 1 wide]</code>: 34</p>
</li>
<li>
<p><strong>Indirect Nodes:</strong> 345 nodes (0.15%)</p>
</li>
</ul>
<hr>
<h3 data-heading="3. Short Interactive Session (Browser, Bouncing Atoms, Image Update)">3. Short Interactive Session (Browser, Bouncing Atoms, Image Update)</h3>
<ul>
<li>
<p><strong>Total Cached Nodes:</strong> 28,839</p>
</li>
<li>
<p><strong>Monomorphic (<code>[1 fast, 0 wide]</code>):</strong> 27,079 nodes (93.89%)</p>
</li>
<li>
<p><strong>Polymorphic Fast Tier (<code>[2..4 fast, 0 wide]</code>):</strong> 1,107 nodes (3.84%)</p>
</li>
<li>
<p><code>[2 fast, 0 wide]</code>: 954 (3.31%) | <code>[3 fast, 0 wide]</code>: 112 | <code>[4 fast, 0 wide]</code>: 41</p>
</li>
<li>
<p><strong>Wide Tier Nodes (<span class="math math-inline">w > 0</span>):</strong> 653 nodes (2.26%)</p>
</li>
<li>
<p><code>[0 fast, 1 wide]</code>: 524 | <code>[0 fast, 2 wide]</code>: 3</p>
</li>
<li>
<p><code>[1 fast, 1 wide]</code>: 100 | <code>[2 fast, 1 wide]</code>: 15 | <code>[3 fast, 1 wide]</code>: 11</p>
</li>
<li>
<p><strong>Indirect Nodes:</strong> 77 nodes (0.27%)</p>
</li>
<li>
<p><strong>Reflective Operations Logged:</strong> 32 calls to <code>allInstancesOf</code>, 1 call to <code>allInstances</code>, and 1 call to <code>pointersBecomeOneWay</code>.</p>
</li>
</ul>
<hr>


Workload | Total Cached Nodes | Monomorphic (1,0) | Polymorphic Fast (2..4,0) | Wide-Tier Containing (w>0) | Indirect Nodes
---- | ---- | ---- | ---- | ---- | ----
AWFY Benchmarks (Avg) | ~1,200 | ~96% | ~3% | < 0.1% | 1–2
Cuis 7.3 Tests | 72,058 | 68,573 (95.16%) | 3,071 (4.26%) | 514 (0.71%) | 187 (0.26%)
Cuis 7.5 Tests | 139,171 | 135,640 (97.46%) | 2,987 (2.15%) | 544 (0.39%) | 183 (0.13%)
Squeak Tests | 228,027 | 220,344 (96.63%) | 5,849 (2.56%) | 1,834 (0.80%) | 345 (0.15%)
Interactive Session | 28,839 | 27,079 (93.89%) | 1,107 (3.84%) | 653 (2.26%) | 77 (0.27%)

Sabine-Schnabeltier: > It's easy to make mistakes and cause memory leaks or weird compilation issues when using raw node operations like `insert()` etc.

You are correct about this and there are multiple issues with the current code. I've reworked it and am testing it in a separate branch.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/AbstractDispatchNode.java
Sabine-Schnabeltier: As we discussed in Slack, we can retain some of the information by sharing the `SqueakObjectClassNode` between the wide nodes and the specialization path. See latest commit.

Sabine-Schnabeltier:
I've just pushed the final commit :crossed_fingers: to wrap up the implementation based on the recent reviews and our Slack discussions.

To address the final pieces of feedback:

* **Raw Node Operations:** I’ve completely reworked `DispatchCache` to use `@Children` arrays instead of manual linked-list node operations, which should resolve the memory leak and compilation concerns.
* **Profiling Info Loss:** As discussed, the `SqueakObjectClassNode` is now shared between the wide nodes and the specialization path. This helps retain profiling information when transitioning from the fast tier to the wide tier.
* **Safety:** Added guards against old deoptimized frames.

Since this PR has evolved quite a bit during review, here is a quick recap of the final, complete state:

1. **Multi-Level Dispatch Cache:** A two-tiered cache (Fast & Wide) with an indirect fallback.
2. **Dispatch Failures (DNU & CI):** Safely handles nested `cannotInterpret:` and implements allocation-free DNU shortcuts (arities 0–3) which are strictly isolated in the fast tier.
3. **Interpreter Optimizations:** Unrolls extended sends (arities 3–5) to prevent allocations and caches selectors independently for reflective calls.

CI results:
https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/30584787525
https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/30592699530
https://github.com/Sabine-Schnabeltier/trufflesqueak/actions/runs/30593903581

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelector0Node.java
```diff
@@ -40,35 +41,93 @@
...
+            }
+
+            // TIER 2: Wide Execution (Class Polymorphism)
+            if (cache.wideNodes.length > 0) {
+                final ClassObject receiverClass = cache.classNode.executeLookup(cache, receiver);
+                final Object lookupResult = getContext().lookup(receiverClass, selector);
+
+                if (lookupResult instanceof CompiledCodeObject targetMethod) {
+                    for (final WideDispatchDataNode<DispatchDirect0Node> currentWide : cache.wideNodes) {
+                        if (currentWide.standardMethod == targetMethod) {
```
fniephaus: 🤖 Wide dispatch no longer checks the direct node assumptions. A `CompiledCodeObject` can invalidate `callTargetStable` while remaining the same object in the method dictionary (for example via `setHeader()`, byte/literal mutation, or `flushCache()`). In that case `lookupResult` is still identical to `standardMethod`, so this path keeps executing the `DirectCallNode`/primitive captured before the modification. The fast tier avoids this via the assumptions stored in each guard. Please also validate `currentWide.dispatchDirectNode.getAssumptions()` here and prune/re-specialize invalid wide entries. The same issue applies to the corresponding wide loops in arities 1–5 and both N-ary dispatchers.

Sabine-Schnabeltier: Great catch! You are absolutely right that we need to handle method invalidations for the Wide nodes.

To fix this and prevent similar differences between tiers, I refactored the caching architecture to use a unified `DispatchEntry` wrapper for both the fast and wide tiers.

Here is an outline of the changes:

Assumption Checking: The fast and wide tiers now share the exact same `DispatchEntry` structure, which natively holds the `Assumption[]` extracted from the direct node. The wide tier evaluation (`isWideCacheHit`) now strictly requires `Assumption.isValidAssumption(assumptions)` alongside the method match.

Automatic Pruning: The `specialize()` method inside `DispatchCacheManager` has been completely rewritten. It now explicitly filters out invalid wide entries (via `isWideValid()`) during cache rebuilding, correctly pruning dead nodes to trigger re-specialization.

Zero-Overhead Promotion: Because both tiers use the unified `DispatchEntry`, promoting a method from the fast tier to the wide tier preserves the direct dispatch node (and its profiling data). We simply drop the fast-tier receiver guards to free memory and move the entry directly into the wide array, completely avoiding costly AST node recreation.

Universal Application: Because the cache manager and entry wrapper are strictly generic and live in `AbstractDispatchNode`, this fix automatically propagates to arities 0–5 and both N-ary dispatchers without duplicating the loop logic in every file.

Sabine-Schnabeltier:
Replaced by #299 and #307

PR #280, opened 2026-07-02
Fix Cuis window sizing on startup and resource leaks

Sabine-Schnabeltier:
Before saving the image, Cuis replaces Display with a tiny bitmap to reduce the image size on disk. Previously, we were saving the window dimensions when the window was resized or when the bitmap was resized. Because of this, when starting up the saved Cuis image, the Display would be tiny. Cuis 7.3 would crash with this small size; Cuis 7.5 would start and you could resize the window manually.

Changed SqueakDisplay to:

- save the window dimensions only when the window is resized
- create the window hidden, showing it when the image finally sets Display dimensions to agree with the image header dimensions

Also, made a couple of thread-safe data synchronization fixes and fixed a potential leak if creating a Cursor failed.


fniephaus (approved):
LGTM, thank you!

PR #281, opened 2026-07-02
ensure thread-safe data sync when registering/deregistering vmThread

Sabine-Schnabeltier:
Refines interrupt registration timing: Moves the VM thread registration to just before the interrupt check. This covers a tiny timing edge case, ensuring that if an interrupt arrives at the exact moment the thread is preparing to park, the unpark signal is guaranteed to be caught.

Streamlines thread deregistration: Centralizes the vmThread = null cleanup by deferring it entirely to the finally block of the executing thread. Prevents a scenario where the interrupt thread could clear the vmThread variable concurrently with the VM thread setting it.


fniephaus (approved):
LGTM, thanks!

PR #282, opened 2026-07-03
Fix scheduler context corruption during multiple simultaneous interrupts

Sabine-Schnabeltier:
**The Bug:**
When multiple semaphores are signaled within a single interrupt check cycle (e.g., a Command+period user interrupt and the same Command+period enqueued as a keyboard event), the interrupt check triggered cascading context switches. Because `TransferToNode` was executed immediately for each awakened process, the `VirtualFrame` of the originally active process was erroneously saved into the `suspendedContext` of several intermediate processes, leading to very odd behavior.

**The Fix:**
This PR introduces a "wake-then-switch" architecture to safely batch interrupt evaluations:

* **Decoupled Waking from Switching:** Introduced `SignalSemaphoreForInterruptNode` to evaluate priority preemptions and mutate semaphore states without touching the Truffle execution frame.
* **State Machine Evaluation:** `CheckForInterruptsNode` now acts as a pure state tracker (`nextActiveProcess` and `nextActiveProcessYields`), folding all pending signals (Interrupt, Timer, Finalization, External) into a single "winning" `Process`.
* **Single Context Switch:** `TransferToNode` is now called exactly once at the end of the interrupt loop, guaranteeing the `VirtualFrame` is safely saved only to the originally interrupted process.
* **Standard Primitives Preserved:** Kept the original `SignalSemaphoreNode` intact so standard bytecode-driven primitives (like `PrimSignal`) retain their required synchronous context-switch behavior.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/io/SqueakDisplay.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/io/SqueakDisplay.java
@@ -912,20 +912,25 @@ public final class SqueakDisplay {
             return;
         }
 
+        final boolean isCommandOrCtrl = (buttons & (KEYBOARD.CTRL | KEYBOARD.CMD)) != 0;
         final int keyChar = toSqueakKey(sdlKeySym);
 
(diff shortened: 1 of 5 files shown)


fniephaus (approved):
LGTM

PR #283, opened 2026-07-06
optionally log screen dimensions and HighDPI flag from image header

Sabine-Schnabeltier:
Added IMAGE FINE logging of the image header's screen dimensions and whether it is asking for a high DPI window.


fniephaus (approved):
LGTM

PR #284, opened 2026-07-06
add download progress output to console

Sabine-Schnabeltier:
Display progress while downloading an image. The output is suppressed completely when the `--quiet` option is supplied. If there is no console, only download completion is output.


fniephaus (commented):
LGTM, thanks!

Issue #286, opened 2026-07-08
DNU when installing Squot


Sabine-Schnabeltier:
Open the TruffleSqueak image, open a workspace, evaluate the following:

`Metacello new
    baseline: 'Squot';
    repository: 'github://hpi-swa/Squot:latest-release/src';
    load.`

after some time, the post-loading step will fail with a DNU:

<img width="1136" height="889" alt="Image" src="https://github.com/user-attachments/assets/a62b5b57-515b-45b6-825b-2cce35c5b70f" />

As you can see in the context inspector, `SquotWorkingCopy` does understand `allSubInstancesDo:`, but manually evaluating the code results in the same error. This seems like some kind of caching problem, but not sure how it could be since this code is being compiled and evaluated "live".

Perhaps something like SquotWorkingCopy starts out Undeclared and moves to Smalltalk when the class is compiled. It appears that the code that declares an undeclared binding does a `becomeForward:`, so this could be due to not following a forwarding pointer, perhaps in `pushLitVar`.

Sabine-Schnabeltier:
If I turn off pointer forwarding here by setting the initial value to `false`, the installation of Squot completes successfully.

https://github.com/hpi-swa/trufflesqueak/blob/6577b033dd471d96dbfe8bf3ab2c3ccdcd9e03c3/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/util/ObjectGraphUtils.java#L250

Sabine-Schnabeltier:
This has been fixed by #288

PR #288, opened 2026-07-10
Follow and resolved pushed literal variables

Sabine-Schnabeltier:
The `value` field of a literal variable `Association` was not being checked to see if it had been forwarded before being pushed onto the stack. See #286 for a failing example case.

This PR resolves the forwarded value and updates the objects involved.

If the literal variable is a member of one of the read-only classes, its value is cached in the `data` array. In this case, we need to update `data` with the resolved value and invalidate to flush the cached value from any JIT produced code.

For the other case, we can simply replace the value field of the literal variable with the resolved value.


fniephaus (approved):
LGTM, thanks!

PR #290, opened 2026-07-12
Use context stack depth to set the Java stack size

Sabine-Schnabeltier:
Links the Java thread stack size to the requested `--context-stack-depth` to prevent Java stack overflow, and unifies headless and UI thread execution.

**Changes**

* **Thread Sizing:** Both headless and UI modes now execute inside a dedicated `TruffleSqueakVM-Thread`. The thread's stack size is allocated as `(contextStackDepth + 512) * 1024` bytes.
* **Centralized Constants:** Moves the default context depth (`2 << 12`) and introduces a new minimum depth (`128`) in `SqueakLanguageConfig.java`.
* **Input Clamping:** Clamps the context stack depth to the minimum in both `TruffleSqueakLauncher` and `SqueakOptions`.
* **Deprecation:** Removes `(0 = disabled)` from the help string since that is no longer supported.

**Results**

With this change, it is now possible to `update code from the server` using the default context stack depth of 8192. This was also tested with a variety of context stack depths down to the minimum of 128.

Also tested headless mode:
 
```
trufflesqueak % $TS_JVM_HOME/bin/trufflesqueak --download-image trufflesqueak --context-stack-depth 35000 -c "Time millisecondsToRun: [30000 factorial]"
[trufflesqueak] Running TruffleSqueak-25.0.1.image on Oracle GraalVM (latency mode)...
[trufflesqueak] Evaluating 'Time millisecondsToRun: [30000 factorial]'...
[trufflesqueak] Result: 986
```

_Note that using smaller context stack depths will result in more Truffle frame flushes and slower code execution._

PR #291, opened 2026-07-14
prevent context depth count from signaling when interrupts are disabled

Sabine-Schnabeltier:
Disable context stack depth overflow detection when disabling interrupts during polyglot, uninterruptible, and internal evaluations. This prevents evaluation failure when the context stack depth is close to the limit and one of these special evaluations is required.

This PR, when used with PR #290, allows Squeak to start up and function normally with a minimum context stack depth of 1.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -89,9 +89,14 @@ import de.hpi.swa.trufflesqueak.util.ObjectGraphUtils;
 @DefaultExpression("get($node)")
 public final class SqueakImageContext {
     private static final ContextReference<SqueakImageContext> REFERENCE = ContextReference.create(SqueakLanguage.class);
+    private static final int SUSPENDED_CONTEXT_STACK_DEPTH = Integer.MIN_VALUE / 2;
 
     /* Encapsulates the state needed to safely execute without interrupts and stack limits */
-    public record SavedExecutionState(boolean insterruptsWereActive, int savedContextDepth) {
+    public record SavedExecutionState(SqueakImageContext context, boolean interruptsWereActive, int savedContextDepth) implements AutoCloseable {
+        @Override
+        public void close() {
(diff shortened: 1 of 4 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
```diff
@@ -292,18 +296,43 @@ public SqueakImage getSqueakImage() {
...
+     *
+     * @return a {@link SavedExecutionState} record containing the original environment state.
+     * This MUST be passed to {@link #resumeNormalExecution(SavedExecutionState)}
+     * inside a {@code finally} block to restore normal VM operations.
+     */
+    public SavedExecutionState suspendNormalExecution() {
+        final boolean wasActive = interrupt.deactivate();
+        final int savedDepth = currentContextStackDepth;
+        currentContextStackDepth = -100_000;
+        return new SavedExecutionState(wasActive, savedDepth);
```
fniephaus: Maybe `SavedExecutionState` could implement `Closable`, so we can use it in a try-with-resources statement?

Something like:
```
        try (var _ = image.suspendNormalExecution()) {
	        // ...
	    }
```

Sabine-Schnabeltier: Great idea!

Sabine-Schnabeltier: Unfortunately, native image doesn't like the try-with-resource shortcut on the fast path and checkstyle doesn't like underscore as a variable name. The code is all set to handle it whenever those restrictions change!

fniephaus: Indeed, try-with-resources statements seem to use `Throwable.addSuppressed()` for exceptions thrown in the block, and that method is blocklisted because it's non-trivial and may allocate an `ArrayList`.

fniephaus (approved):
LGTM, thanks!

PR #292, opened 2026-07-15
follow forwarding pointer when tracing specialObjectsArray

Sabine-Schnabeltier:
Following the instructions for creating the [TruffleSqueak Test Image](https://github.com/hpi-swa/trufflesqueak/blob/main/docs/images.md#trufflesqueak-test-image-creation) leads to an interpreter crash.

It appears that the `specialObjectsArray` is replaced using `becomeForward:`. This PR ensures that the forwarded `specialObjectsArray` is traced.


Sabine-Schnabeltier:
I think the tests are failing because an incorrect `image` submodule has been included in the PR.

PR #293, opened 2026-07-16
Display scaling and default language

Sabine-Schnabeltier:
This PR changes the `Polyglot` desktop buttons so that they resize properly when opened on a high-DPI screen. It also prompts for the default `Polyglot` language so that the displayed workspace (and subsequent workspaces opened by the desktop button) have a language selected.

**Recommended:** If the TruffleSqueak image is created on a `displayScale = 1.0` screen it will look good when opened on any DPI screen.

If the `displayScale = 2.0` when the image is created, the text in the desktop buttons will be small when displayed on a `displayScale = 1.0` screen. (The font sizing complication is due to `StrikeFont` having a small set of fixed font sizes to choose from.)


Sabine-Schnabeltier:
Converted to draft while some bugs are ironed out...

PR #295, opened 2026-07-21
Upgrade to SDL `3.4.12` and initialize bindings at build-time

fniephaus:
This PR updates TruffleSqueak to use SDL 3.4.12 via LWJGL 3.4.2. It also splits reachability metadata into separate files that accompany each binding. I also used the opportunity to simplify Windows support.

Most importantly, I customized `jextract` to emit bindings that can be initialized at build-time, which improves FFM API performance significantly. In particular, FFM API and the bindings heavily make use of `MethodHandle`s and they must be initialized at build-time to be fast when TruffleSqueak is compiled into a native image. To analyze this, I built TruffleSqueak with the following options to dump the compiler graphs for `SDL_free`:

```
NATIVE_IMAGE_OPTIONS="-H:+TrackNodeSourcePosition -H:MethodFilter="*SDL_free*" -H:Dump=:3" mx --env native build
```

When looking at the "after phase PhaseSuite" graphs (last but one entry in igv), we can see an invoke of `MethodHandle.invokeBasic`, which is currently slow on Native Image as it runs in an interpreter:

Screenshot 2026-07-21 at 19 41 25

The regenerated bindings use unbounded [downcall handles that can be build-time initialized](https://github.com/oracle/graal/pull/13309), which allows Native Image to identify and (indirectly) call the C function:

Screenshot 2026-07-21 at 19 42 05

While customizing jextract, I also used the opportunity to disable `TRACE_DOWNCALLS` at build-time, so the corresponding checks and invokes are also removed from the compilation units:

Screenshot 2026-07-21 at 19 40 44

I will update the instructions for regenerating the bindings with my custom jextract soon.


src/de.hpi.swa.trufflesqueak.interpreterproxy.bindings/src/de/hpi/swa/trufflesqueak/interpreterproxy/bindings/sqVirtualMachine_h$shared.java
```diff
@@ -27,12 +27,23 @@ public class sqVirtualMachine_h$shared {
     public static final ValueLayout.OfDouble C_DOUBLE = (ValueLayout.OfDouble) Linker.nativeLinker().canonicalLayouts().get("double");
     public static final AddressLayout C_POINTER = ((AddressLayout) Linker.nativeLinker().canonicalLayouts().get("void*"))
             .withTargetLayout(MemoryLayout.sequenceLayout(java.lang.Long.MAX_VALUE, C_CHAR));
-    // Windows uses LLP64, so C `long` is 32-bit there even on 64-bit builds.
-    public static final ValueLayout.OfLong C_LONG /* sqInt */ = System.getProperty("os.name").toLowerCase().contains("win")
-                    ? sqVirtualMachine_h.C_LONG_LONG
-                    : (ValueLayout.OfLong) Linker.nativeLinker().canonicalLayouts().get("long");
+    public static final ValueLayout C_LONG = (ValueLayout) Linker.nativeLinker().canonicalLayouts().get("long");
```
Sabine-Schnabeltier, line 30: I think this will result in a 32-bit layout on Windows unless the linker has been customized to use LP64 instead of LLP64 on Windows. The comment that has been removed implies that `C_LONG` should be 64 bits.

If it needs to be 64 bits, could you use:

`public static final ValueLayout C_INT64 = ValueLayout.JAVA_LONG;`

I couldn't easily find any references to `C_LONG`, so the point may be moot.

Sabine-Schnabeltier (commented):
Excellent! (I assume you saw the Codacy comments about the unneeded `import` and a couple of `finals`.)

src/de.hpi.swa.trufflesqueak.interpreterproxy.bindings/src/de/hpi/swa/trufflesqueak/interpreterproxy/bindings/sqVirtualMachine_h$shared.java
fniephaus: `C_LONG` is no longer used because I hardcoded `typedef int64_t sqInt;` in `sqMemoryAccess.h`, which is always 64 bit. `jextract` turns that into `C_LONG_LONG` instead which is guaranteed to be at least 64 bits.

Do you think it's better if we use your suggested `C_INT64`?

fniephaus:
> I assume you saw the Codacy comments about the unneeded import and a couple of finals.

I did but I didn't want to customize jextract further. I think it's ok if generated code throws some warnings.

src/de.hpi.swa.trufflesqueak.interpreterproxy.bindings/src/de/hpi/swa/trufflesqueak/interpreterproxy/bindings/sqVirtualMachine_h$shared.java
fniephaus: My agent concluded:
> If you want maximum semantic clarity, C_INT64 would better document “exactly 64-bit integer”, but C_LONG_LONG is ABI-correct on the relevant platforms.

PR #296, opened 2026-07-28
Guard for forwarded class in super-send dispatch

Sabine-Schnabeltier:
During `processBytecode()`, `InterpreterSistaV1Node` and `InterpreterV3PlusClosuresNode` cache the method's class in a `DispatchSuperNaryNode` when processing an `EXT_SEND_SUPER` byte code.

If that cached class is mutated during execution, the class will be forwarded to its replacement but the `DispatchSuperNaryNode` will continue to dispatch to the pre-mutation superclass.

This PR adds a guard to ensure that the cached class has not been forwarded. If the class is forwarded, the node is replaced with a new node that caches the forwarded class.

The following code snippet demonstrates the problem and fails on `main` and works correctly with the changes in this PR.

```
| superA superB subClass instance result |

"1. Create the Old Superclass (returns 1)"
superA := Object subclass: #SuperSendTestA 
    instanceVariableNames: '' 
    classVariableNames: '' 
    poolDictionaries: '' 
    category: 'TruffleSqueak-Tests'.
superA compile: 'dummy: a with: b with: c ^ 1'.

"2. Create the New Superclass (returns 2)"
superB := Object subclass: #SuperSendTestB
    instanceVariableNames: '' 
    classVariableNames: '' 
    poolDictionaries: '' 
    category: 'TruffleSqueak-Tests'.
superB compile: 'dummy: a with: b with: c ^ 2'.

"3. Create our Target Class, initially inheriting from SuperA"
subClass := superA subclass: #SuperSendTestSub 
    instanceVariableNames: '' 
    classVariableNames: '' 
    poolDictionaries: '' 
    category: 'TruffleSqueak-Tests'.

"4. The Time Bomb: A method that changes its own superclass mid-execution"
subClass compile: 'doMutatingSuperSend
    "Redefine the class to inherit from SuperB. This triggers a becomeForward:"
    SuperSendTestB subclass: #SuperSendTestSub 
        instanceVariableNames: '''' 
        classVariableNames: '''' 
        poolDictionaries: '''' 
        category: ''TruffleSqueak-Tests''.
        
    "Execute extended super send. 
     - Unpatched AST will use the stale class and call SuperA >> dummy (returns 1).
     - Patched AST will heal itself, follow the forwarding pointer to the new class, 
       and call SuperB >> dummy (returns 2)."
    ^ super dummy: 1 with: 2 with: 3'.

"5. Execute the test"
instance := subClass new.
result := instance doMutatingSuperSend.

"6. Cleanup"
subClass removeFromSystem.
superA removeFromSystem.
superB removeFromSystem.

"7. Assert the fix"
self assert: result = 2.
^ 'Success: The AST healed and returned ', result printString
```

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
@@ -9,7 +9,6 @@ package de.hpi.swa.trufflesqueak.model;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
 
-import com.oracle.truffle.api.dsl.Idempotent;
 import org.graalvm.collections.UnmodifiableEconomicMap;
 
 import com.oracle.truffle.api.CompilerAsserts;
@@ -196,7 +195,6 @@ public abstract class AbstractSqueakObjectWithHash extends AbstractSqueakObject 
         squeakHashAndBits |= FORWARDED_BIT;
     }
 
-    @Idempotent
     public final boolean isNotForwarded() {
         return (squeakHashAndBits & FORWARDED_BIT) == 0;
     }
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
@@ -9,6 +9,7 @@ package de.hpi.swa.trufflesqueak.model;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
 
+import com.oracle.truffle.api.dsl.Idempotent;
 import org.graalvm.collections.UnmodifiableEconomicMap;
 
 import com.oracle.truffle.api.CompilerAsserts;
@@ -195,6 +196,7 @@ public abstract class AbstractSqueakObjectWithHash extends AbstractSqueakObject 
         squeakHashAndBits |= FORWARDED_BIT;
     }
 
+    @Idempotent
     public final boolean isNotForwarded() {
         return (squeakHashAndBits & FORWARDED_BIT) == 0;
     }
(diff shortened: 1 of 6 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -99,11 +100,22 @@ public abstract static class DispatchSuperNaryNode extends AbstractDispatchNaryN
             this.methodClass = methodClass;
         }
 
-        @Specialization(assumptions = {"methodClass.getClassHierarchyAndMethodDictStable()", "dispatchDirectNode.getAssumptions()"})
```
fniephaus, line 102: In `ClassObject.forwardTo()`, the first assumption is invalidated, which should cause the cached entry to be removed and replaced with a new entry automatically. Not quite sure why that doesn't work here yet.

Sabine-Schnabeltier: You are right! The reason that the assumption was not catching this failure is that `#becomeForward:` is sent prior to the super-send, so when the `DispatchSuperNaryNode` is executed, it creates the `DispatchDirectNaryNode` cache using the stale class pointer.

I've removed the guard and changed the cache creation expression so that it follows both the class pointer and the superclass pointer.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
```diff
@@ -195,6 +196,7 @@ protected final void setForwardedBit() {
         squeakHashAndBits |= FORWARDED_BIT;
     }
 
+    @Idempotent
```
fniephaus: This is actually not true, the return value of `isNotForwarded()` can change. Reverting this.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
fniephaus: Yes, that's much better. However, I still didn't like that we keep following forwarding pointers, so I refactored this some more: now, the node caches the method and we call `getMethodClassSlow()` which resolves any forwarding pointers. This also allows us to reduce duplication.

fniephaus:
I also added your test via https://github.com/hpi-swa/trufflesqueak/commit/fdef140edd979e10b97cc28eb3accf54100bce88.

fniephaus (approved):
LGTM, thanks for this critical bug fix!

Sabine-Schnabeltier:
Thanks for your improvements and for incorporating the test!

PR #298, opened 2026-08-03
ephemeronsQueue is now accessed by more than one thread

Sabine-Schnabeltier:
The first commit is definitely needed; the second may be too stringent. What I was seeing is that the tests were timing out and the semaphore had ~50000 excess signals.

I am still seeing the too many open files errors locally.

Later: I realize now that the test that prevents the signaling of the semaphore should be moved to the test in `image`, prior to the poll of the weak references queue. I can do this tomorrow if you haven't already finished your PR before then.


fniephaus:
Thanks!

PR #299, opened 2026-08-11
Refine dispatch fallbacks (DNU/CI) and optimize extended sends

Sabine-Schnabeltier:
This is a fresh PR to replace the previous drafts (#252 and #275) with a much cleaner, two-commit history. I've set aside the multi-level dispatch implementation for now and focused this PR on the fallback paths.

Here are the main improvements:

**Allocation-Free DNU Shortcuts**: For missing methods with 3 or fewer arguments, we can now skip allocating a temporary `Array` and `Message` object. If configured (via `primitiveSetDNUShortcutSelectors`), a failed send transforms directly into a call to a shortcut selector, simply tacking the original selector on as the final argument.

**Zero-Overhead when Disabled**: We introduced a Truffle Assumption (`dnuShortcutsAbsent`) to ensure that if an image doesn't configure these DNU shortcuts, the VM incurs zero overhead checking for them.

**Nested `#cannotInterpret:`**: We now gracefully handle cascading dispatch failures. If a method dictionary is `nil` and `#cannotInterpret:` itself fails, we build nested `Message` objects to preserve the exact context of the original failure all the way down the chain.

**Smarter Cache Invalidation**: DNU fallback nodes now track an image-global "absent selector" assumption. If a developer compiles a method that was previously missing, the AST instantly invalidates the old DNU fallback nodes so they can link to the newly compiled method.

**Unrolled Extended Sends**: We unrolled the extended send bytecodes for arities 0–5 in the interpreters (`InterpreterSistaV1Node` and `InterpreterV3PlusClosuresNode`). This eliminates expensive `popN()` array allocations when running heavily on the interpreter path.

**`perform:` Cache Fix**: Cleaned up the `PrimPerform` primitives by stripping out duplicate receiver guards. Caching is now fully delegated down to the dispatch nodes, which prevents an AST cache explosion for highly polymorphic reflective calls.

### Testing

Verified that DNU shortcuts for arities 0-3 successfully bypass standard `Message` allocations.

Verified that corrupting multiple method dictionaries correctly triggers the nested #cannotInterpret: logic.

[TruffleSqueakDispatchFallbackTests.st.zip](https://github.com/user-attachments/files/30918871/TruffleSqueakDispatchFallbackTests.st.zip)

(Note: Replaces draft PRs #252 and #275)

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
@@ -82,10 +82,11 @@ public final class DispatchSelectorNaryNode extends AbstractDispatchSelectorNode
             return DispatchSelectorNaryNodeFactory.DispatchNaryNodeGen.create(selector);
         }
 
-        @Specialization(guards = "guard.check(receiver)", assumptions = "dispatchDirectNode.getAssumptions()", limit = "INLINE_METHOD_CACHE_LIMIT")
+        @Specialization(guards = {"guard.check(receiver)", "arguments.length == cachedArity"}, assumptions = "dispatchDirectNode.getAssumptions()", limit = "INLINE_METHOD_CACHE_LIMIT")
         protected static final Object doDirect(final VirtualFrame frame, final Object receiver, final Object[] arguments,
                         @SuppressWarnings("unused") @Cached("create(receiver)") final LookupClassGuard guard,
-                        @Cached("create(selector, guard, arguments.length)") final DispatchDirectNaryNode dispatchDirectNode) {
+                        @SuppressWarnings("unused") @Cached("arguments.length") final int cachedArity,
+                        @Cached("create(selector, guard, cachedArity)") final DispatchDirectNaryNode dispatchDirectNode) {
             return dispatchDirectNode.execute(frame, receiver, arguments);
         }
 
@@ -109,10 +110,11 @@ public final class DispatchSelectorNaryNode extends AbstractDispatchSelectorNode
             return DispatchSelectorNaryNodeFactory.DispatchPerformNaryNodeGen.create(selector);
         }
 
-        @Specialization(guards = "guard.check(receiver)", assumptions = "dispatchDirectNode.getAssumptions()", limit = "INLINE_METHOD_CACHE_LIMIT")
+        @Specialization(guards = {"guard.check(receiver)", "arguments.length == cachedArity"}, assumptions = "dispatchDirectNode.getAssumptions()", limit = "INLINE_METHOD_CACHE_LIMIT")
         protected static final Object doDirect(final VirtualFrame frame, final Object receiver, final Object[] arguments,
                         @SuppressWarnings("unused") @Cached("create(receiver)") final LookupClassGuard guard,
-                        @Cached("create(selector, guard, arguments.length)") final DispatchDirectNaryNode dispatchDirectNode) {
+                        @SuppressWarnings("unused") @Cached("arguments.length") final int cachedArity,
+                        @Cached("create(selector, guard, cachedArity)") final DispatchDirectNaryNode dispatchDirectNode) {
             return dispatchDirectNode.executeWithCheckedArguments(frame, receiver, arguments);
         }
 
@@ -122,7 +124,7 @@ public final class DispatchSelectorNaryNode extends AbstractDispatchSelectorNode
         @SuppressWarnings("truffle-static-method")
         protected final Object doIndirect(final VirtualFrame frame, final Object receiver, final Object[] arguments,
                         @Cached final DispatchIndirectNaryNode dispatchNode) {
-            return dispatchNode.execute(frame, false, selector, receiver, arguments);
+            return dispatchNode.execute(frame, true, selector, receiver, arguments);
         }
     }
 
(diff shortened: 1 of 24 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
```diff
@@ -131,12 +137,11 @@ public final class ControlPrimitives extends AbstractPrimitiveFactoryHolder {
...
     protected abstract static class PrimPerform0Node extends AbstractPrimitiveWithFrameNode implements Primitive1WithFallback {
         @SuppressWarnings("unused")
-        @Specialization(guards = {"selector == cachedSelector", "guard.check(receiver)"}, assumptions = "dispatchDirectNode.getAssumptions()", limit = "PERFORM_SELECTOR_CACHE_LIMIT")
+        @Specialization(guards = "selector == cachedSelector", limit = "PERFORM_SELECTOR_CACHE_LIMIT")
         protected static final Object perform0Cached(final VirtualFrame frame, final Object receiver, final NativeObject selector,
                         @Cached("selector") final NativeObject cachedSelector,
-                        @Cached("create(receiver)") final LookupClassGuard guard,
-                        @Cached("create(cachedSelector, guard, true)") final DispatchDirect0Node dispatchDirectNode) {
-            return dispatchDirectNode.execute(frame, receiver);
+                        @Cached("create(cachedSelector)") final Dispatch0Node dispatchNode) {
```
fniephaus: 🤖 [P1] This generic dispatch node loses `perform:`'s primitive-failure semantics after its receiver PIC becomes megamorphic. `Dispatch0Node.doIndirect()` hard-codes `canPrimFail = false` (and the same is true for arities 1–5), whereas this primitive's old direct node and the outer megamorphic specialization use `true`. After enough receiver classes, performing a selector with the wrong argument count therefore stops raising `PrimitiveFailed.BAD_NUMBER_OF_ARGUMENTS`; with assertions off, `ResolveMethodNode` returns the mismatched method and it can be invoked with the wrong frame shape. Please use a perform-specific wrapper whose indirect path passes `true`, or propagate `canPrimFail` through these nodes.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -91,6 +99,33 @@ protected final Object doIndirect(final VirtualFrame frame, final Object receive
...
+            return dispatchDirectNode.executeWithCheckedArguments(frame, receiver, arguments);
+        }
+
+        @ReportPolymorphism.Megamorphic
+        @Specialization(replaces = "doDirect")
+        @HostCompilerDirectives.InliningCutoff
+        @SuppressWarnings("truffle-static-method")
+        protected final Object doIndirect(final VirtualFrame frame, final Object receiver, final Object[] arguments,
+                        @Cached final DispatchIndirectNaryNode dispatchNode) {
+            return dispatchNode.execute(frame, false, selector, receiver, arguments);
```
fniephaus: 🤖 [P1] `DispatchPerformNaryNode` must preserve the checked `perform:withArguments:` behavior when its receiver cache becomes megamorphic. Passing `false` here bypasses the bad-argument failure in `ResolveMethodNode.doMethod`; this should pass `true`, matching `PrimPerformWithArguments2Node`'s outer megamorphic path and this class's `executeWithCheckedArguments()` direct path.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -91,6 +99,33 @@ protected final Object doIndirect(final VirtualFrame frame, final Object receive
...
+
+        @NeverDefault
+        public static DispatchPerformNaryNode create(final NativeObject selector) {
+            return DispatchSelectorNaryNodeFactory.DispatchPerformNaryNodeGen.create(selector);
+        }
+
+        @Specialization(guards = "guard.check(receiver)", assumptions = "dispatchDirectNode.getAssumptions()", limit = "INLINE_METHOD_CACHE_LIMIT")
+        protected static final Object doDirect(final VirtualFrame frame, final Object receiver, final Object[] arguments,
+                        @SuppressWarnings("unused") @Cached("create(receiver)") final LookupClassGuard guard,
+                        @Cached("create(selector, guard, arguments.length)") final DispatchDirectNaryNode dispatchDirectNode) {
```
fniephaus: 🤖 [P1] The cached direct node depends on `arguments.length`, but that value is not part of the specialization guard. Reusing this node for the same receiver class and selector with a different `perform:withArguments:` array is especially unsafe for a missing selector with DNU shortcuts: `DispatchDirectShortcutFallbackNaryNode.checkNumArguments()` is a no-op, so with assertions disabled a shorter array throws `ArrayIndexOutOfBoundsException`, while a longer array silently drops trailing arguments and invokes the shortcut selected for the first arity. Please cache the arity separately and add `arguments.length == cachedArity` to the guard.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
```diff
@@ -513,6 +563,55 @@ public Object execute(final VirtualFrame frame, final Object receiver, final Obj
...
+            this.arity = arity;
+        }
+
+        @Override
+        @ExplodeLoop
+        public Object execute(final VirtualFrame frame, final Object receiver, final Object[] arguments) {
+            assert arguments.length == arity : "Shortcut arity mismatch!";
+
+            // Build the shortcut arguments: [arg0, arg1, ..., argN, selector]
+            final Object[] shortcutArgs = new Object[arity + 1];
```
fniephaus: 🤖 [P3] Since shortcuts are capped at arity 3, this intermediate `shortcutArgs` array and subsequent copy in `FrameAccess.newWith(..., shortcutArgs)` can be avoided with an arity switch that calls the fixed `FrameAccess.newWith` overloads directly, as the fixed-arity shortcut nodes already do. The indirect shortcut frame builder below has the same opportunity. This would reduce interpreter-path allocation/copying for reflective DNU shortcuts.

fniephaus (commented):
🤖 I found three correctness issues in the new reflective-dispatch caching, plus one small allocation improvement for the N-ary shortcut path. The project builds successfully with GraalVM 25.1.3/javac.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/primitives/impl/ControlPrimitives.java
Sabine-Schnabeltier: Thanks for catching this!

Added `canPrimFail` to the fixed-arity dispatch nodes to propagate to the megamorphic paths.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
Sabine-Schnabeltier: Changed to `true`

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
Sabine-Schnabeltier: Done! Thanks again!

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/dispatch/DispatchSelectorNaryNode.java
Sabine-Schnabeltier: Excellent suggestion! Done!

PR #300, opened 2026-08-11
add Squeak 6.1 image; fix image read of non-integer Context stack pointers

Sabine-Schnabeltier:
OSVM internally tolerates non-integer stack pointers in `Contexts`, returning `0` for non-integer values. However, Squeak 6.1 images contain `Context` instances with `nil` stack pointers when read from disk.

Since Smalltalk can only read or write integers from/to the stack pointer, converting `nil` or non-integer stack pointers to `0` during the image read phase preserves expected semantics and prevents downstream runtime issues.

Changes in this PR:

- Converts non-integer stack pointer values to 0 when reading the image file.
- Adds the Squeak 6.1 image to the list of run-time selectable images.

_Note: Squeak 6.1 images do not have the high-DPI bit set, so the display will look quantized on high-DPI monitors. See #301 _


fniephaus (approved):
LGTM

Issue #273, opened 2026-06-23
Add support for Pharo


kilian-kier:
TruffleSqueak currently supports Squeak (Cuis) images only. [Pharo](https://pharo.org/) is a fork of Squeak but with differences in its kernel class hierarchy, block closure system, FFI implementation, and many more that require targeted adaptations throughout the VM. Supporting Pharo would expand TruffleSqueak's user base and allow Pharo developers to benefit from the GraalVM/Truffle infrastructure.

### Todos

- [x] SOA: Pharo does not populate specific element in the special object array which are assumed by TruffleSqueak to be there
- [ ] Image Loading Adaptation: for example Pharo has `MetaclassForTraits` as a second subclass (subclass of subclass) of `Metaclass`. This breaks TruffleSqueak's current assumption during image loading that only a single metaclass subclass exists.
- [ ] Block Closures: Pharo introduces `CleanBlockClosure`, `ConstantBlockClosure`, `ConstantBlockClosure(1/2/3)Arg` which probably must be handled in the VM
- [ ] Compiled Method Format: Pharo uses a fixed 5-byte source pointer trailer
- [ ] FFI: Pharo replaced the old SqueakFFI with ThreadedFFI
- [ ] Primitives/plugins: implement missing essential primitives and plugins, such as FileAttributesPlugin.
- [ ] Compiler: Pharo uses `OpalCompiler` instead of the traditional `Compiler`
- [ ] Semantic differences: 
  - String comparison primitive 158 returns different result encoding (raw 1/2/3 vs. -1/0/1)
  - Method dictionaries can contain WideString selectors (not just ByteString)
  - Support primitives with indices greater than 578.

### Open Questions

- FFI Native Memory Lifecycle: How should primitives like primitiveGetAddressOfOOP be handled, which assume the VM allocates and manages the underlying native memory?
- Dialect Detection: What is the most robust way to determine whether the loaded image is Squeak/Cuis or Pharo? (This is crucial for conditionally branching behavior regarding e.g. Compiled Method Format and semantic differences).
- Is it feasible to maintain it without forking the repo?

fniephaus:
Thanks for opening this ticket and for looking into supporting Pharo images in TruffleSqueak!

Going through your open questions:

> **FFI Native Memory Lifecycle:** How should primitives like `primitiveGetAddressOfOOP` be handled, which assume the VM allocates and manages the underlying native memory?

[...]

PR #302, opened 2026-08-12
Generalize class object fill-in to support Pharo and Squeak images

kilian-kier:
This PR is relate to [#273](https://github.com/hpi-swa/trufflesqueak/issues/273).

Refactors `fillInClassObjects` to dynamically discover `Metaclass` and all subclasses (e.g. Pharo's TraitedMetaclass and MetaclassForTraits) removing Squeak-specific class hierarchy assumptions. Replaced direct Metaclass equality checks with full-depth hierarchy traversal (`isMetaclassChunk`)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageReader.java
```diff
@@ -332,77 +333,117 @@ private void initObjects() {
...
                 final SqueakImageChunk classChunk = chunkMap.get(potentialClassPtr);
-                if (classChunk.getSqueakClass() == image.metaClass) {
-                    /* Derive classIndex from current position in class table. */
-                    highestKnownClassIndex = p << SqueakImageConstants.CLASS_TABLE_MAJOR_INDEX_SHIFT | i;
+                if (classChunk == null || classChunk.isNil()) {
+                    continue;
+                }
+
+                if (isMetaclassChunk(classChunk, sqMetaclass)) {
                     assert classChunk.getWordSize() == METACLASS.INST_SIZE;
```
fniephaus: 🤖 This assertion still assumes every metaclass object has exactly the six slots defined by `METACLASS`. The condition above now also matches instances of metaclass subclasses, but Pharo’s `TraitedMetaclass` adds four instance variables and `MetaclassForTraits` adds another one. Loading such an image with assertions enabled will therefore abort here. Could this instead verify only that `METACLASS.THIS_CLASS` is in bounds (or accept sizes greater than or equal to `METACLASS.INST_SIZE`)?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageReader.java
```diff
@@ -332,77 +333,117 @@ private void initObjects() {
...
+                    continue;
+                }
+
+                if (isMetaclassChunk(classChunk, sqMetaclass)) {
+                    final SqueakImageChunk classInstanceChunk = getThisClassChunk(classChunk);
+                    if (classInstanceChunk != null) {
+                        final ClassObject classObject = classInstanceChunk.asClassObject();
+                        if (classObject != null) {
+                            classObject.fillin(classInstanceChunk);
+                            if (behaviorClasses.contains(classObject.getSuperclassOrNull())) {
```
fniephaus: 🤖 This propagation is dependent on class-table order: if a deeper `ClassDescription` subclass appears before its superclass, the superclass is not yet in `behaviorClasses`, and the subclass is never revisited. Class-table order is not guaranteed to be topological. Consider filling all classes first and then using a recursive/memoized ancestor check or a fixed-point traversal.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageReader.java
```diff
@@ -332,77 +333,117 @@ private void initObjects() {
...
-                    assert classObject != null;
-                    classObject.fillin(classInstance);
-                    if (inst.contains(classObject.getSuperclassOrNull())) {
-                        inst.add(classObject);
-                        classObject.setInstancesAreClasses();
+                if (classChunk == null || classChunk.isNil()) {
+                    continue;
+                }
+
+                if (isMetaclassChunk(classChunk, sqMetaclass)) {
```
fniephaus: 🤖 Performance suggestion: `isMetaclassChunk()` walks the hierarchy for every class-table entry here after pass 1 already performed the same scan and hierarchy checks. Collect the matching metaclass/class-instance chunks during pass 1 and reuse that collection in pass 2; this would eliminate a full table scan and the repeated hierarchy traversals.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageReader.java
kilian-kier: Relaxed the assertion to be greater than or equal to `METACLASS.INST_SIZE`

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageReader.java
kilian-kier: Solved with a 3rd pass and `checkAndMarkBehaviorClass`. Class loading still doesn't seem to take noticeably longer

Sabine-Schnabeltier (changes requested):
The overall approach for handling Pharo's extended metaclass hierarchy makes sense, particularly the memoization logic you introduced for `checkAndMarkBehaviorClass`.

I have a couple of suggestions to improve the performance and reduce memory allocations in `fillInClassObjects`.

## 1. Memoizing the Metaclass Check

Right now, the code walks the entire inheritance chain for every class table entry. This can be optimized by using a recursive method that memoizes the intermediate classes, mirroring what you did in `checkAndMarkBehaviorClass`. If you pass the chunk's class directly, this can be reduced to a single method.

You can add these as instance variables:

```java
private final HashSet<SqueakImageChunk> knownMetaclassClasses = new HashSet<>();
private final HashSet<SqueakImageChunk> knownNonMetaclassClasses = new HashSet<>();

```

And replace the metaclass check with this recursive method:

```java
private boolean inheritsFromMetaclass(final SqueakImageChunk chunkClass, final SqueakImageChunk sqMetaclass) {
    if (chunkClass == null || chunkClass.isNil()) {
        return false;
    }
    if (chunkClass == sqMetaclass || knownMetaclassClasses.contains(chunkClass)) {
        return true;
    }
    if (knownNonMetaclassClasses.contains(chunkClass)) {
        return false;
    }

    if (inheritsFromMetaclass(getSuperclassChunk(chunkClass), sqMetaclass)) {
        knownMetaclassClasses.add(chunkClass);
        return true;
    } else {
        knownNonMetaclassClasses.add(chunkClass);
        return false;
    }
}

```

## 2. Merging Pass 1 and Pass 2

You could avoid allocating the `metaclassChunks` list entirely. There's no need to wait for all metaclasses to be found before starting to extract their `thisClass` instances. You could combine the first two loops:

```java
final List<ClassObject> classInstances = new ArrayList<>();

// Combined Pass 1 & 2
for (int p = 0; p < SqueakImageConstants.CLASS_TABLE_ROOT_SLOTS; p++) {
    // ... (standard table traversal) ...
        
        if (inheritsFromMetaclass(classChunk.getClassChunk(), sqMetaclass)) {
            assert classChunk.getWordSize() >= METACLASS.INST_SIZE;
            final ClassObject metaClassObject = classChunk.asClassObject();
            assert metaClassObject != null;
            metaClassObject.setInstancesAreClasses();

            final SqueakImageChunk classInstanceChunk = getThisClassChunk(classChunk);
            if (classInstanceChunk != null) {
                final ClassObject classObject = classInstanceChunk.asClassObject();
                if (classObject != null) {
                    classObject.fillin(classInstanceChunk);
                    classInstances.add(classObject);
                }
            }
        }
    // ...
}

```

Pass 3 remains the same, using the populated `classInstances` list.

Finally, what are your thoughts on testing this? I realize TruffleSqueak might not be ready to actually run a Pharo image yet, but would it make sense to include a basic "load-only" test with a minimal Pharo image just to verify these changes to `SqueakImageReader` correctly parses the hierarchy?

fniephaus:
@kilian-kier would you like to incorporate @Sabine-Schnabeltier's suggestions still, or show we do it in another PR?

> Finally, what are your thoughts on testing this?

While it would indeed be great to test loading a Pharo image, it seems more work is needed to load all objects:

```
Caused by: org.graalvm.polyglot.PolyglotException: java.lang.ClassCastException: class de.hpi.swa.trufflesqueak.model.VariablePointersObject cannot be cast to class de.hpi.swa.trufflesqueak.model.BlockClosureObject (de.hpi.swa.trufflesqueak.model.VariablePointersObject and de.hpi.swa.trufflesqueak.model.BlockClosureObject are in module de.hpi.swa.trufflesqueak of loader 'app')
        at de.hpi.swa.trufflesqueak.model.ContextObject.fillin(ContextObject.java:103)
        at de.hpi.swa.trufflesqueak.image.SqueakImageReader.fillInObjects(SqueakImageReader.java:472)
        at de.hpi.swa.trufflesqueak.image.SqueakImageReader.initObjects(SqueakImageReader.java:330)
        at de.hpi.swa.trufflesqueak.image.SqueakImageReader.run(SqueakImageReader.java:83)
        at de.hpi.swa.trufflesqueak.image.SqueakImageReader.load(SqueakImageReader.java:64)
        at de.hpi.swa.trufflesqueak.SqueakImage$SqueakImageNode.execute(SqueakImage.java:40)
```

PR #303, opened 2026-08-13
Fix race condition in squeakHashAndBits

kilian-kier:
This PR fixes a race-condition in the usage of squeakHashAndBits. Even with the existence of HASH_AND_BITS_HANDLE for atomic accesses, it was not used everywhere.

MarkBitsConcurrencyTest will mostly fail without those changes

### DIFF
--- a/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/MarkBitsConcurrencyTest.java
+++ b/src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/MarkBitsConcurrencyTest.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright (c) 2026 Software Architecture Group, Hasso Plattner Institute
+ * Copyright (c) 2026 Oracle and/or its affiliates
+ *
+ * Licensed under the MIT License.
+ */
+package de.hpi.swa.trufflesqueak.test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.junit.Test;
+
+import de.hpi.swa.trufflesqueak.model.AbstractSqueakObjectWithHash;
+import de.hpi.swa.trufflesqueak.model.ClassObject;
+import de.hpi.swa.trufflesqueak.model.PointersObject;
+
+@SuppressWarnings("static-method")
+public final class MarkBitsConcurrencyTest extends AbstractSqueakTestCaseWithDummyImage {
+    private static final long FLAG_ITERATIONS = 10_000_000;
+    private static final int MARK_OBJECT_COUNT = 20_000;
+    private static final int MARK_THREAD_COUNT = 4;
+
+    /**
+     * Verifies that concurrent bit-field updates (boolean flags vs. GC mark bits) are thread-safe.
+     * One thread toggles boolean flags while another marks/unmarks the object; neither may lose
+     * updates or mutate the identity hash.
+     */
+    @Test
+    public void testFlagAndMarkingBitsDoNotLoseUpdates() throws InterruptedException {
+        final PointersObject object = instantiate(createFreshTestClass());
+        final int hash = 0x2A2A2A;
+        object.setSqueakHash(hash);
+        object.setBooleanDBit(); /* Set once, must never be observed unset again. */
+
+        final Queue<String> failures = new ConcurrentLinkedQueue<>();
+        final CountDownLatch start = new CountDownLatch(1);
+
+        final Thread booleanBitsThread = new Thread(() -> {
+            try {
+                start.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            for (long i = 0; i < FLAG_ITERATIONS; i++) {
+                object.setBooleanABit();
+                if (!object.isBooleanASet()) {
+                    failures.add("Lost setBooleanABit in iteration " + i);
+                    return;
+                }
+                object.clearBooleanABit();
+                if (object.isBooleanASet()) {
+                    failures.add("Lost clearBooleanABit in iteration " + i);
+                    return;
+                }
+                if (!object.isBooleanDSet()) {
+                    failures.add("Lost the untouched boolean D bit in iteration " + i);
+                    return;
+                }
+                if (object.getSqueakHashInt() != hash) {
+                    failures.add("Hash changed to " + object.getSqueakHashInt() + " in iteration " + i);
+                    return;
+                }
+            }
+        }, "booleanBitsThread");
+
+        final Thread markingBitsThread = new Thread(() -> {
+            try {
+                start.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            for (long i = 0; i < FLAG_ITERATIONS; i++) {
+                if (!object.tryToMarkWith(true)) {
+                    failures.add("Could not mark in iteration " + i);
+                    return;
+                }
+                if (!object.isMarkedWith(true)) {
+                    failures.add("Lost the mark in iteration " + i);
+                    return;
+                }
+                object.unmarkWith(true);
+                if (object.isMarkedWith(true)) {
+                    failures.add("Lost the unmark in iteration " + i);
+                    return;
+                }
+            }
+        }, "markingBitsThread");
+
+        booleanBitsThread.start();
+        markingBitsThread.start();
+        start.countDown();
+        booleanBitsThread.join();
+        markingBitsThread.join();
+
+        assertEquals("Lost updates: " + failures, 0, failures.size());
+        assertEquals("Hash survived", hash, object.getSqueakHashInt());
+        assertTrue("Boolean D bit survived", object.isBooleanDSet());
+    }
+
+    /**
+     * Verifies that marking objects is atomic across concurrent threads.
+     * When multiple threads race to mark the same object, {@code tryToMarkWith(true)} must return
+     * {@code true} for exactly 1 thread and {@code false} for all others. Across N objects,
+     * the total number of successful marks across all threads must equal N.
+     */
+    @Test
+    public void testConcurrentMarkingMarksEachObjectExactlyOnce() throws InterruptedException {
+        final ClassObject testClass = createFreshTestClass();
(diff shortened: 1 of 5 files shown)


Sabine-Schnabeltier:
Hi Killian,

Could you please run the TruffleSqueak Tests Action on this PR branch on your fork and post a link to the results here so we can check for any test regressions? (Since automated CI runs on PRs are disabled here).

Thanks!

kilian-kier:
Hi Sabine,

here's the link: https://github.com/kilian-kier/trufflesqueak/actions/runs/31778394291

src/de.hpi.swa.trufflesqueak.test/src/de/hpi/swa/trufflesqueak/test/MarkBitsConcurrencyTest.java
```diff
@@ -0,0 +1,169 @@
...
+                        "booleanBitsThread");
+
+        final Thread markingBitsThread = new Thread(
+                        () -> runMarkingBitsWorker(object, failures, start),
+                        "markingBitsThread");
+
+        booleanBitsThread.start();
+        markingBitsThread.start();
+        start.countDown();
+        booleanBitsThread.join();
```
fniephaus, line 58: 🤖 JUnit does not propagate uncaught exceptions from child threads. If either worker throws unexpectedly, these `join()` calls can complete and the test can still pass; in this first test, the final hash/D-bit assertions may remain true even if a worker exited immediately. Could these workers be submitted as `Future`s and checked with `get()`, or could each `Throwable` be captured and rethrown after joining? The same concern applies to the worker threads in the second test.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
```diff
@@ -89,8 +91,10 @@ protected AbstractSqueakObjectWithHash(final SqueakImageChunk chunk) {
 
     @SuppressWarnings("this-escape")
     protected AbstractSqueakObjectWithHash(final AbstractSqueakObjectWithHash original) {
-        /* Preserves flags and resets hash. */
-        squeakHashAndBits = original.squeakHashAndBits & SQUEAK_HASH_FLAGS_MASK;
+        /*
+         * Preserves flags and resets hash and mark bits.
+         */
+        squeakHashAndBits = original.squeakHashAndBits & SQUEAK_HASH_FLAGS_MASK & ~MARKING_MASK;
```
fniephaus, line 97: 🤖 This is still a plain read of the shared field, despite the new access contract requiring synchronization through `HASH_AND_BITS_HANDLE`. If a shallow copy overlaps an update to the original, it can observe stale Boolean/forwarding flags. Please read the original through `HASH_AND_BITS_HANDLE.getAcquire(original)` before applying the masks; the direct assignment to the new, unpublished object can remain plain.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
```diff
@@ -63,7 +65,7 @@ public abstract class AbstractSqueakObjectWithHash extends AbstractSqueakObject
     /* Generate new hash if hash is 0 (see SpurMemoryManager>>#hashBitsOf:). */
     protected static final int HASH_UNINITIALIZED = 0;
 
-    private int squeakHashAndBits;
+    @ThreadAccess(value = Mode.CAS, orderedBy = "HASH_AND_BITS_HANDLE") private int squeakHashAndBits;
```
fniephaus, line 68: 🤖 `Mode.CAS` is documented as meaning that the field is updated only by `compareAndSet`, but several setters below use atomic bitwise RMW operations. Since this annotation is intended to document the access contract, could it use a broader mode such as `ATOMIC_RMW`, or could the mode documentation explicitly cover both CAS and VarHandle atomic RMW operations?

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
```diff
@@ -136,7 +140,7 @@ public long getSqueakHash() {
 
     public final int getSqueakHashInt() {
         assert assertNotForwarded();
-        return squeakHashAndBits >>> SQUEAK_HASH_SHIFT;
+        return ((int) HASH_AND_BITS_HANDLE.getAcquire(this)) >>> SQUEAK_HASH_SHIFT;
```
fniephaus, line 143: 🤖 Potential performance improvement: every hash and Boolean-bit read now uses `getAcquire()`. These reads appear to inspect only packed scalar state, without consuming any separately published object data. If that invariant is intentional, `getOpaque()` may provide the required atomic/coherent access while avoiding acquire barriers, especially on ARM; the retry loops may similarly be candidates for release-mode weak CAS. Could we benchmark these weaker access modes and document whether acquire ordering is actually required?

fniephaus (commented):
🤖 I found three review concerns and one potential performance improvement.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/model/AbstractSqueakObjectWithHash.java
fniephaus: 🤖 Follow-up: since `ThreadAccess` has `SOURCE` retention, no annotation processor/static-analysis rule consumes it, and this is currently its only use, it does not provide any functional enforcement. Could you drop `ThreadAccess` from this PR and document the contract directly on the field instead? For example:

```java
/**
 * Packed identity hash and flags. Once published, access this field only through
 * {@link #HASH_AND_BITS_HANDLE}. Use atomic RMW operations for individual flags
 * and CAS loops for whole-field updates. Plain access is permitted only while
 * constructing an unpublished object.
 */
private int squeakHashAndBits;
```

An annotation would make sense later if the project applies it consistently and adds tooling that verifies the declared access mode.

Sabine-Schnabeltier:
Thank you for this PR and the substantial effort clearly behind it!

You are right that in a standard multi-threaded environment, mixing plain bitwise assignments with `VarHandle` CAS loops on the same field would cause lost updates.

However, because of TruffleSqueak's specific execution model, we actually do not need to protect the boolean bits or hash initialization with atomic operations, and doing so results in a severe performance regression (~30%).

Here is a breakdown of why the current implementation is safe across all of TruffleSqueak's threads:

* **The Main Thread blocks during Object Graph Traversal:** The only time background threads interact with Squeak objects is during graph traversal (e.g., ephemeron processing or `allInstances`). When this happens, the main interpreter thread submits the tasks in `ObjectGraphUtils.runTasks()` and immediately calls `Future.get()`. The main thread is halted until all background threads finish. Because the main thread is parked, it cannot execute code that triggers `setBooleanABit()` or `setSqueakHash()` at the same time a background thread is calling `tryToMarkWith()`.

* **External Threads (UI & Interrupts) do not touch Objects:**
* The `SqueakDisplay` thread runs an SDL3 event loop, but it never mutates Squeak objects directly. Instead, it places primitive event data into a thread-safe `ConcurrentLinkedDeque` (the `deferredEvents` queue). The main thread eventually polls this queue and performs the actual object updates.
* Similarly, the background thread in `CheckForInterruptsState` only flips `SHOULD_TRIGGER` flag and unparks the main VM thread. It queues up signals (like `semaphoresToSignal`) for the main thread to process safely, leaving object headers untouched.

* **GraalVM Performance Constraints:** Because the background traversal threads only contend with *each other* (which is why `tryToMarkWith` requires the CAS loop), we can keep standard bitwise operations (like `squeakHashAndBits |= BOOLEAN_A_BIT`) for the other flags. This allows the Graal compiler to easily inline and optimize them during Partial Evaluation. If we switch these to `getAndBitwiseOrRelease` or CAS loops, it would force hardware-level memory barriers on the single-threaded fast path, which could heavily penalize the JIT compiler.

TruffleSqueak relies on this intentional lack of synchronization for performance during normal execution, so, while making the header flags thread safe is a good idea in theory, in practice  it is not actually necessary and is exacting a performance price.

I am curious whether some variation of your `MarkBitsConcurrencyTest` could be added to the test suite to specifically test the concurrent marking logic of `tryToMarkWith` between background threads (excluding concurrent boolean bits and hash manipulations). What do you think?

Sabine-Schnabeltier (commented):
Thank you for this PR and the substantial effort clearly behind it!

You are right that in a standard multi-threaded environment, mixing plain bitwise assignments with `VarHandle` CAS loops on the same field would cause lost updates.

However, because of TruffleSqueak's specific execution model, we actually do not need to protect the boolean bits or hash initialization with atomic operations, and doing so results in a severe performance regression (~30%).

Here is a breakdown of why the current implementation is safe across all of TruffleSqueak's threads:

* **The Main Thread blocks during Object Graph Traversal:** The only time background threads interact with Squeak objects is during graph traversal (e.g., ephemeron processing or `allInstances`). When this happens, the main interpreter thread submits the tasks in `ObjectGraphUtils.runTasks()` and immediately calls `Future.get()`. The main thread is halted until all background threads finish. Because the main thread is parked, it cannot execute code that triggers `setBooleanABit()` or `setSqueakHash()` at the same time a background thread is calling `tryToMarkWith()`.

* **External Threads (UI & Interrupts) do not touch Objects:**
* The `SqueakDisplay` thread runs an SDL3 event loop, but it never mutates Squeak objects directly. Instead, it places primitive event data into a thread-safe `ConcurrentLinkedDeque` (the `deferredEvents` queue). The main thread eventually polls this queue and performs the actual object updates.
* Similarly, the background thread in `CheckForInterruptsState` only flips `SHOULD_TRIGGER` flag and unparks the main VM thread. It queues up signals (like `semaphoresToSignal`) for the main thread to process safely, leaving object headers untouched.

* **GraalVM Performance Constraints:** Because the background traversal threads only contend with *each other* (which is why `tryToMarkWith` requires the CAS loop), we can keep standard bitwise operations (like `squeakHashAndBits |= BOOLEAN_A_BIT`) for the other flags. This allows the Graal compiler to easily inline and optimize them during Partial Evaluation. If we switch these to `getAndBitwiseOrRelease` or CAS loops, it would force hardware-level memory barriers on the single-threaded fast path, which could heavily penalize the JIT compiler.

TruffleSqueak relies on this intentional lack of synchronization for performance during normal execution, so, while making the header flags thread safe is a good idea in theory, in practice  it is not actually necessary and is exacting a performance price.

I am curious whether some variation of your `MarkBitsConcurrencyTest` could be added to the test suite to specifically test the concurrent marking logic of `tryToMarkWith` between background threads (excluding concurrent boolean bits and hash manipulations). What do you think?

kilian-kier:
Makes sense, thanks for the explanation. I will close this PR now and try to rework `MarkBitsConcurrencyTest` for testing the concurrent marking logic

Issue #163, opened 2022-05-25
`UnsupportedSpecializationException`: Unexpected values provided for `ContextObjectWriteNodeGen` when creating `Context` instances manually


LinqLover:
Practical example to reproduce:

- Install/get an image with [SimulationStudio](https://github.com/LinqLover/SimulationStudio) (for instance, the image from [this bundle](https://github.com/hpi-swa-lab/squeak-tracedebugger/releases/tag/untagged-20220512-072554) should do it)
- Do it:
   ```smalltalk
   Simulator evaluate: [1 + 2]
   ```
- Expected behavior: Answers `3`
- Actual behavior: crash (`UnsupportedSpecializationException`: Unexpected values provided for `ContextObjectWriteNodeGen`)

Minimal example to debug:

- Do it:
   ```smalltalk
   thisContext privSender: Object new.
   ```
- Expected behavior: A `cannotReturn:` is put on the stack (and subsequently, the current project is emergency-exited because the debugger cannot display the stack, but the VM keeps alive - see OSVM for reference)
- Actual behavior: VM crashes, same error as above

---

It would be nice if TruffleSqueak could be more robust with regard to unusual contexts - just like the OpenSmalltalk VM, which does not raise an error unless the interpreter actually attempts to return to an invalid sender on the context.

For some background: SimulationStudio is relying on the possibility to put arbitrary objects (subinstances of `Context`) on the stack to modify the behavior of the image-side simulator while reusing the `Context` interface. These objects are never actually interpreted by the VM, but the simulation protocol on `Context` can handle them like regular contexts.

fniephaus:
Thanks for raising the issue! So far, TruffleSqueak is quite constrained as to what you can do with `Contexts` because they are very performance-critical. Historically, we only added support for the things usually used by Squeak/Smalltalk and its tests and intentionally sprinkled bits and pieces that help us find new cases that need to be supported: e.g., there are lots of assertions and, as you found out, tight specializations that trigger `UnsupportedSpecializationExceptions`.

In this particular case, TruffleSqueak only allows setting the sender to another `Context` object or `nil`:
https://github.com/hpi-swa/trufflesqueak/blob/a068eca13ca09ac1de50bc4572dd53333fab8718/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/accessing/ContextObjectNodes.java#L79-L88

[...]

PR #304, opened 2026-08-15
Force Context mutations to the heap

Sabine-Schnabeltier:
This PR is the first step towards addressing issue #163. It updates the interpreter to ensure that any modification to a `Context`'s instance variables strictly occurs on the heap. By moving the context safety checks out of individual variable store nodes and into a centralized execution boundary, the VM preemptively protects active execution states without incurring per-store overhead.

### **Key Changes**

* **Static Context Store Detection:** The `processBytecode` method now analyzes bytecodes to detect potential stores to context instance variables (targeting indices smaller than `CONTEXT.INST_SIZE`). This sets a new `modifiesContext` flag on `AbstractInterpreterNode`.
* **Context Class Validation:** A new `couldMutateContextVariables` method confirms whether the executing method belongs to the `Context` class hierarchy.
* **Preemptive Heap Flushing:** When an interpreter node begins execution, if `modifiesContext` is true and the receiver is a `ContextObject`, the VM explicitly triggers `ensureContextIsNotActive` to ensure the receiver is on the heap.
* **Code Cleanup:** Redundant `checkForAndHandlePCModification` and `FrameAccess.externalizePCAndSP` calls have been removed from individual store bytecodes.
* **Casting Fixes:** Minor adjustments were made to `getMethodClassAssociation` and its callers to remove overly strict `AbstractPointersObject` casts.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
@@ -37,6 +37,7 @@ import de.hpi.swa.trufflesqueak.model.AbstractSqueakObject;
 import de.hpi.swa.trufflesqueak.model.AbstractSqueakObjectWithClassAndHash;
 import de.hpi.swa.trufflesqueak.model.ArrayObject;
 import de.hpi.swa.trufflesqueak.model.BlockClosureObject;
+import de.hpi.swa.trufflesqueak.model.ClassObject;
 import de.hpi.swa.trufflesqueak.model.CompiledCodeObject;
 import de.hpi.swa.trufflesqueak.model.ContextObject;
 import de.hpi.swa.trufflesqueak.model.layout.ObjectLayouts.ASSOCIATION;
@@ -76,6 +77,7 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
 
     protected final CompiledCodeObject code;
     protected final boolean isBlock;
+    protected final boolean modifiesContext;
 
     @CompilationFinal(dimensions = 1) private final Object[] data;
     @CompilationFinal(dimensions = 1) private final byte[] profiles;
@@ -89,7 +91,8 @@ public abstract class AbstractInterpreterNode extends AbstractInterpreterInstrum
         final int endPC = code.getMaxPCZeroBased();
         data = new Object[endPC];
         profiles = new byte[endPC];
-        processBytecode(startPC, endPC);
+        final boolean hasPotentialContextIVStore = processBytecode(startPC, endPC);
+        modifiesContext = couldMutateContextVariables(hasPotentialContextIVStore);
     }
 
     @SuppressWarnings("this-escape")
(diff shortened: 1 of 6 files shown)


src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/AbstractInterpreterNode.java
```diff
@@ -102,15 +105,56 @@ public AbstractInterpreterNode(final AbstractInterpreterNode original) {
...
+            // Fast-path rejection: must have between 1 and Context's number of instance variables
+            final ClassObject contextClass = getContext().methodContextClass;
+            final int methodClassInstSize = methodClass.getBasicInstanceSize();
+
+            if (methodClassInstSize > 0 && methodClassInstSize <= contextClass.getBasicInstanceSize()) {
+                // Walk the superclass chain until we hit a class with 0 instance variables (e.g., Object)
+                ClassObject current = contextClass;
+                do {
+                    if (methodClass == current) {
+                        return true;
```
fniephaus, line 149: So here we check whether the executing method belongs to the Context class hierarchy.

I'm not sure that is sufficient because primitive methods never cause a new call if the primitive was successful: Any method could do, for example, `thisContext stackp: 2` which modifies the currently executing context without a message send.

Thus, any method could modify context vars which is why we had to move the checks close to the store operation.

Sabine-Schnabeltier: Fabio, that is a really good point about primitives bypassing the new call activation!

However, I think that scenario is actually unaffected by this PR, since this PR targets bytecode execution (e.g., store receiver variable). Because bytecodes can only store into `self`, a method can only use a bytecode to mutate a `Context` if it belongs to the `Context` class hierarchy. That means the preemptive `couldMutateContextVariables` check gives us the exact same coverage as the old method of checking every individual store bytecode.

For the `thisContext stackp: 2` example, that hits Primitive 76. If the primitive succeeds, it bypasses the bytecode loop entirely and is handled directly by `PrimStoreStackPointerNode`. Because it completely skips the bytecode loop, it would have bypassed the old checks inside the individual variable store nodes anyway.

Does that make sense, or do you think there is an edge case I'm missing where a non-Context method could somehow force a bytecode store into a `Context`? Possibly the not-yet-implemented experimental Sista bytecodes?

fniephaus: > bytecodes can only store into `self`. ... That means the preemptive `couldMutateContextVariables` check gives us the exact same coverage as the old method of checking every individual store bytecode.

You are right, of course! Great that we can replace the overhead per store receiver variable with a simple per execution check.

> Possibly the not-yet-implemented experimental Sista bytecodes?

I don't think so. IIRC they are more or less specialized versions of the existing bytecodes.

src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/nodes/interpreter/InterpreterV3PlusClosuresNode.java
```diff
@@ -114,6 +120,9 @@ protected void processBytecode(final int startPC, final int endPC) {
                     switch (variableType(descriptor)) {
                         case 0: {
                             setData(currentPC, insert(SqueakObjectAt0NodeGen.create()));
+                            if (variableIndex < CONTEXT.INST_SIZE) {
+                                hasPotentialContextIVStore = true;
```
fniephaus: Dropping these changes as this is a push bytecode.

fniephaus (approved):
LGTM

fniephaus (approved):
LGTM

fniephaus (approved):
LGTM

Sabine-Schnabeltier:
Thanks!

PR #305, opened 2026-08-18
Introduce ContextProxy

Sabine-Schnabeltier:
**Depends on #304** 

This PR builds on the recent heap-based `Context` changes by introducing `ContextProxy` to safely manage structurally invalid or reflectively mutated contexts.

Changes:

- Introduces the `ContextProxy` state, featuring automatic dematerialization of Truffle frames upon incompatible reflective writes, and dynamic materialization when a valid execution state resumes.
- Centralizes context boundary checks in `ContextObject`, enforcing physical Truffle frame capacity over logical Squeak limits.
- Consolidates out-of-bounds temporary variable and stack accesses into safe no-ops across frames and proxies, preventing indexing exceptions.
- Reorders `ContextObject` initialization in `SqueakImageContext.getDoItContextNode()` to properly establish code objects before receivers during image boot.

With these changes, all previously failing examples in #163 are correctly functioning (see image below).

ContextProxy tests

For maximum compatibility with OSVM, we should consider a command line option to disable the optimization (trimming) of `Context` stack frames imposed by the current execution model.

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -428,8 +428,8 @@ public final class SqueakImageContext {
         final CompiledCodeObject doItMethod = (CompiledCodeObject) methodNode.send(this, "generate");
 
         final ContextObject doItContext = new ContextObject(doItMethod.getSqueakContextSize());
-        doItContext.setReceiver(NilObject.SINGLETON);
         doItContext.setCodeObject(doItMethod);
+        doItContext.setReceiver(NilObject.SINGLETON);
         doItContext.setInstructionPointer(0);
         doItContext.setStackPointer(doItMethod.getNumTemps());
         doItContext.setSenderUnsafe(NilObject.SINGLETON);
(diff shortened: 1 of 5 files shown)


fniephaus:
> For maximum compatibility with OSVM, we should consider a command line option to disable the optimization (trimming) of Context stack frames imposed by the current execution model.

What do you mean by this? Why can't we maximize compatibility by default?

Sabine-Schnabeltier:
What I meant is that OSVM always allocates fixed-size contexts (16 to 56 slots) based on the method header flag. Squeak code may sometimes reflectively rely on this by writing into the unused 'dead space' of those slots. TruffleSqueak, on the other hand, trims `VirtualFrame` allocations to the maximum stack depth required by the method's bytecodes. This PR handles the difference by ignoring out-of-bounds dead-space writes, which prevents crashes but technically diverges from OSVM's literal memory layout.

If we forced TruffleSqueak to perfectly emulate OSVM by default, we would have to allocate 16 to 56 slots for every single method activation. This would likely decrease Truffle's peak performance, but I am not certain.

A command-line flag (e.g., `--osvm-frame-sizes`) would just give users an escape hatch to perfectly emulate OSVM's memory layout at the cost of performance, which might be useful for highly reflective sandbox libraries like `SimulationStudio`.

fniephaus:
An alternative solution that still provides compatibility involves falling back to auxiliary frame slots when "dead space" is accessed. Doing so would cause some performance overheads but maybe that's a good trade off as these kind of accesses are rather uncommon, and not recommended (as you'd rely on VM internal behavior).

Sabine-Schnabeltier:
I think it would be better to keep the implementation simple for the common case. If someone is advanced enough to be taking over the stack manipulation, it would be no problem for them to disable the stack size optimization with a command line flag.

PR #307, opened 2026-09-09
Multilevel dispatch reloaded

Sabine-Schnabeltier:
This PR replaces the Truffle DSL-generated polymorphic inline caches for message dispatch (`Dispatch0Node` through `DispatchNaryNode`) with a custom Multi-Tier Dispatch Architecture.

On the AWFY benchmarks, the Interpreter performance is faster than `main` and the Peak performance is slightly faster than `main`. **Bounce**, **List** and **NBody** are slower and also have a larger Tier 2 code size which could indicate that the Smalltalk method inlining may be too aggressive for those benchmarks. Perhaps tuning of the inlining budget could better balance the tradeoff between code size and execution speed with respect to instruction cache sizes.

Local testing (Mac mini M2, 16 GB, macOS 15.7.9) of the Cuis 7.3 and 7.5 test suite runtimes show an improvement of performance by 5% and 2%, respectively.

### Core Architecture

- **Tier 0 (Monomorphic)**: Direct execution using a single fast receiver check (`LookupClassGuard`), bypassing array iterations entirely.
- **Tier 1 (Fast / Receiver-Based Polymorphism)**: Checked dispatch up to `DISPATCH_CACHE_LIMIT` (4), where each entry can guard up to `LOOKUP_CACHE_LIMIT` (8) receiver types mapping to the same method target.
- **Tier 2 (Wide / Target-Based Polymorphism)**: Target-based dispatching based on resolved method targets rather than receiver types.
- **Tier 3 (Indirect / Megamorphic)**: Fallback indirect execution.

---

_This PR replaces #275 and is built upon PR #299 and will be rebased once #299 is accepted._

### DIFF
--- a/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
+++ b/src/de.hpi.swa.trufflesqueak/src/de/hpi/swa/trufflesqueak/image/SqueakImageContext.java
@@ -166,7 +166,8 @@ public final class SqueakImageContext {
     // The maximum message arity that supports DNU shortcuts.
     public static final int MAX_DNU_SHORTCUT_ARITY = 3;
 
-    @CompilationFinal(dimensions = 1) private NativeObject[] dnuShortcutSelectors = new NativeObject[0];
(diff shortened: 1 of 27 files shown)


Sabine-Schnabeltier:
The first four commits overlap with #299

